PR #97, opened 2026-04-12
fix: allow webdav connect when root propfind returns 405

### LINKED ISSUE #95: [BUG] NextCloud connection - 405 Method Not Allowed
**Describe the bug**
I am trying to add NextCloud to the server list and connect to it using WebDAV. No matter what I do, I always receive "405 Method Not Allowed."

WebDAV connection is working on other clients, so I have no idea what to do here or where to check

**To Reproduce**
Steps to reproduce:
1. create new NC connection
2. connect to the server

**Expected behavior**
I would expect a successful connection

**Screenshots**

Image

**Environment:**
 - OS: Windows 11
 - AeroFTP Version: 3.4.8
 - NextCloud version: 32.0.8 (the latest)

saschabuehrle:
Fixes #95

Some WebDAV servers return 405 for PROPFIND / even though WebDAV is available. The connect handshake now falls back to OPTIONS when the initial PROPFIND gets 405, so valid endpoints can still connect.

Greetings, saschabuehrle

### DIFF
--- a/src-tauri/src/providers/webdav.rs
+++ b/src-tauri/src/providers/webdav.rs
@@ -1178,6 +1178,33 @@ impl StorageProvider for WebDavProvider {
             StatusCode::FORBIDDEN => Err(ProviderError::AuthenticationFailed(
                 "Invalid credentials".to_string(),
             )),
+            StatusCode::METHOD_NOT_ALLOWED => {
+                tracing::debug!(
+                    "[WebDAV] PROPFIND / returned 405, retrying with OPTIONS for capability check"
+                );
+
+                let options_response = self
+                    .request(Method::OPTIONS, "/")
+                    .send()
+                    .await
+                    .map_err(|e| ProviderError::ConnectionFailed(e.to_string()))?;
+
+                let options_status = options_response.status();
+                if options_status.is_success() {
+                    self.connected = true;
+                    if let Some(ref initial_path) = self.config.initial_path {
+                        if !initial_path.is_empty() {
+                            self.current_path = initial_path.clone();
+                        }
+                    }
+                    Ok(())
+                } else {
+                    Err(ProviderError::ConnectionFailed(format!(
+                        "Server returned status: {}",
+                        options_status
+                    )))
+                }
+            }
             status => Err(ProviderError::ConnectionFailed(format!(
                 "Server returned status: {}",
                 status


axpnet:
Hi, thanks a lot! I'm away for the weekend, I'll check everything tomorrow 🙏🏼👍🏽

axpnet:
Thanks for the contribution, Sascha! The PROPFIND 405 fallback is a solid fix and handles a real-world edge case nicely. Glad to have it merged. Welcome aboard, and feel free to open more PRs if you spot anything else.

axpnet:
Hi @saschabuehrle,

apologies for resurfacing a pull request from April. Your fix is still in the codebase and still doing its job: the OPTIONS fallback is now the last resort after a well known path probe, and it is what lets people connect to servers that block PROPFIND on the root. Thank you for it, it was a real bug and you fixed it properly.

I am writing about licensing housekeeping, and I would rather ask you openly than quietly work around it.

I was going through who holds copyright in this repository, and the answer turned out to be short. Apart from my own commits and the dependency bot, yours is the only contribution in it, and about ten of your lines are still live in the WebDAV provider. That means any future licensing decision about AeroFTP depends on two people rather than one, and I would rather settle that now, while the commit is recent and you are easy to reach, than try to track someone down in three years.

AeroFTP is GPL-3.0 and I intend to keep it that way. It is free software, it stays free software, and I have no plan to change that. What I am asking for is narrower and duller than it may sound: permission to license your contribution under terms in addition to GPL-3.0, so that if the project ever does need to adjust its licensing, the decision is not blocked. To be straight with you rather than vague, that includes the possibility of offering a commercial licence to someone who cannot use copyleft. I think that is unlikely and I am not working towards it, but I would rather tell you it is on the table than leave it out and have you find out later.

I am not asking you to transfer or give up your copyright. It stays yours, and your contribution stays in the history with your name on it either way.

If you are happy with this, a reply on this thread is enough. Something like:

> I, Sascha Buehrle, confirm that my contribution to axpdev-lab/aeroftp in PR #97 may be licensed by the project under terms in addition to GPL-3.0.

If you would rather not, that is completely fine and please do say so. Nothing bad happens: the code stays exactly where it is under the GPL, and I will work around it on my side. I would rather have a clear no than an awkward silence.

Either way, thank you for the fix and for the clean pull request.

Alessandro (axpnet)

PR #163, opened 2026-05-05
ci(snap): bump snapcore/action-build to v2 to clear LXD-OOM

axpnet:
## Summary

Bumps `snapcore/action-build` from v1 (`3bdaa03e`, 2024) to v2 to clear the silent LXD-OOM that killed the Snap build on v3.6.9, v3.6.10, v3.7.0 and v3.7.1.

v2 ships LXD 5.21 stable with a fixed memory allocator. Inputs and outputs are unchanged.

`continue-on-error: true` is removed: a real Snap build failure now blocks the release like the rest of the Linux pipeline (deb/rpm/AppImage), instead of letting Snap Store silently fall behind.

## Test plan
- [ ] First tag after merge (v3.7.2 or later) cuts a Snap build that completes
- [ ] Snap Store stable channel advances past 3.6.10

## Summary by CodeRabbit

* **Chores**
  * Updated the Linux Snap packaging build configuration to use the latest build action version and enforce stricter failure handling in the release pipeline. Build failures will now prevent incomplete releases.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -169,18 +169,16 @@ jobs:
           echo "AppImage re-packaged: ${ORIG_SIZE}MB -> ${NEW_SIZE}MB (system libs only)"
 
       # Build Snap package using Snapcraft (Tauri doesn't generate .snap)
-      # NOTE 2026-04-28: snapcore/action-build@3bdaa03e is currently dying
-      # silently during LXD setup on ubuntu-22.04 runners (3 reruns of v3.6.9
-      # all silent-OOM at the same point). Marked continue-on-error so the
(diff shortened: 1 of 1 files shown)


axpnet:
Closing: this PR is based on a wrong premise. `snapcore/action-build@v2` does not exist. The action has been frozen at v1.3.0 (`3bdaa03e`, 2024-07-15) for almost two years, and `canonical/action-build` is a sync of the same commit. There is no upstream fix to bump to.

The real cause of the silent LXD-OOM is the `ubuntu-22.04` runner itself. Tracking issue follows so the fix is not lost: extract the Snap build into its own job on `ubuntu-24.04` (LXD 5.21+ native), keeping the main matrix on `22.04` for AppImage glibc compatibility.

PR #323, opened 2026-06-15
fix(ftp): include dotfiles in recursive delete enumeration

axpnet:
## Problem

`rm -r` on an FTP directory that contains a dotfile (e.g. an AeroCrypt overlay's `.aeroftp-crypt.json`, or `.env` / `.gitignore`) fails with `550 Remove directory operation failed`. The recursive sweep deletes every visible entry, then the final `RMD` fails because the still-present invisible dotfile makes the directory non-empty.

Surfaced during AeroCrypt overlay testing; it is a pre-existing plain-FTP limitation, not an overlay bug. SFTP and WebDAV already return dotfiles, so FTP is the outlier.

## Root cause (two layers)

1. On servers that do not advertise MLSD (notably vsftpd, which does not implement it), `list_inner` falls back to a bare `LIST`, and vsftpd hides dotfiles on a bare `LIST`. So `rmdir_recursive` never sees the dotfile, leaves it, and the final RMD hits a non-empty directory.
2. `parse_dos_listing` was too permissive (`parts[2].parse().unwrap_or(0)`). The `LIST -a` output adds `.` and `..` rows; `parse_unix_listing` correctly rejects them, but `parse_listing` then fell through to `parse_dos_listing`, which parsed a Unix `.` row into a bogus entry like `"ftp 4096 Jun 13 15:17 ."`. That made recursive delete issue `DELE .` and 550 on ordinary populated directories. This is why an earlier naive `LIST -a <path>` attempt regressed.

## Fix

- New `list_inner_opts(path, include_hidden)`; `list_inner` becomes a wrapper passing `false`. With `include_hidden = true`: the MLSD path is unchanged (it already enumerates dotfiles, filtering only `.`/`..`); the `LIST` fallback does `CWD <dir>` + bare `LIST -a` (the portable form; the combined `LIST -a <path>` is server-specific) and restores the working directory best-effort.
- New `list_for_delete` (mirrors `list()`'s stale-connection retry); `rmdir_recursive` uses it. `include_hidden` is opt-in to recursive delete only; public `list()` semantics are unchanged (GUI / sync / ls unaffected).
- Harden `parse_dos_listing` to require `parts[2]` be `<DIR>` or numeric, so it no longer resurrects the `.`/`..` rows that `LIST -a` adds. No regression on real IIS DOS listings.

## Tests

- New unit test `list_a_output_keeps_dotfiles_and_drops_dot_dirs`.
- Gate green: `cargo fmt --check`, `cargo clippy --lib -D warnings`, `cargo clippy --bin aeroftp-cli --tests -D warnings`, ftp unit tests 10/10.
- Live-verified on a Docker vsftpd: `rm -r` on a dir with root + nested dotfiles + `.aeroftp-crypt.json` now exits 0 and removes everything; ordinary populated dir still exits 0; `ls` (no `--all`) unchanged.

## Known limitation (out of scope)

On FTP, `ls --all` still does not surface dotfiles (public `list()` does not use `-a`). Deliberately not changed here to avoid touching GUI/sync/ls semantics.

Co-Authored-By: EhudKirsh <EhudKirsh@users.noreply.github.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved FTP directory listing to properly identify and display hidden files across different server types
  * Fixed recursive directory deletion to reliably remove hidden files left by some FTP servers
  * Enhanced FTP response parsing to prevent misinterpretation of server output formats

* **Tests**
  * Added verification for hidden file handling in directory listings

### DIFF
--- a/src-tauri/src/providers/ftp.rs
+++ b/src-tauri/src/providers/ftp.rs
@@ -370,11 +370,19 @@ impl FtpProvider {
             return None;
         }
 
+        // In the DOS format parts[2] is always either "<DIR>" or the numeric
+        // size. Requiring that keeps this parser from resurrecting a Unix line
+        // that parse_unix_listing already rejected (e.g. the "." / ".." rows that
+        // `LIST -a` adds, whose parts[2] is the group name): without this guard
+        // those become bogus entries and recursive delete would issue `DELE .`.
         let is_dir = parts[2] == "<DIR>";
         let size: u64 = if is_dir {
             0
         } else {
-            parts[2].parse().unwrap_or(0)
+            match parts[2].parse() {
+                Ok(value) => value,
+                Err(_) => return None,
+            }
         };
         let name = parts[3..].join(" ");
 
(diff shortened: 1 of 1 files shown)


axpnet:
Superseded: this fix is already on main and ships in v4.0.5 (dcc000f70) - recursive delete now enumerates dotfiles. Closing the PR.

Issue #465, opened 2026-07-26
snap: consume graphics-core22 instead of bundling Mesa drivers


axpnet:
## Problem

Snapcraft's GPU linter reports that the AeroFTP core22 snap primes its own Mesa DRI/EGL/GLX/libdrm stack. The current package contains `usr/lib/x86_64-linux-gnu/dri/*_dri.so`, `libEGL_mesa`, `libGLX_mesa`, `libgbm`, and vendor `libdrm` libraries even though hardware-specific userspace should come from a graphics content provider.

Canonical's current guidance recommends the `graphics-core22` content interface with `mesa-core22`, the maintained wrapper, and cleanup helper. This matters beyond package size: shipping a private Mesa stack can conflict with host/NVIDIA userspace and was visible while diagnosing the confined WebKit GUI regression in #462 / PR #463.

Official guide: https://ubuntu.com/frame/docs/22/how-to/use-snap-graphics/

## Acceptance criteria

- [ ] Add a `graphics-core22` content plug targeting `$SNAP/graphics`, default provider `mesa-core22`.
- [ ] Run the maintained `graphics-core22-wrapper` in the app command chain without breaking the GNOME extension's desktop launcher.
- [ ] Add the official `graphics-core22-cleanup` part after AeroFTP and remove provider-owned Mesa/DRI/GLVND/libdrm files from the primed app snap.
- [ ] Add the documented X11/libdrm layouts required by the wrapper.
- [ ] Confirm `snapcraft lint` no longer emits `gpu:` warnings for bundled drivers; assess remaining library-linter warnings individually (dynamic loading vs genuinely unused) rather than blanket-ignore them.
- [ ] CI: G1 ABI, G2 installed CLI payload, and G3 confined GUI readiness under Xvfb all pass.
- [ ] Validate on a real GPU-capable disposable CI/VM path as a separate acceleration check; do not use the owner's desktop session.
- [ ] Do not publish to the Snap Store until the migration and GUI gates are green.

## Scope note

PR #463 should first restore a renderable confined GUI and prove startup. This issue owns the larger graphics-provider migration so it can be reviewed and validated independently rather than combining two runtime architecture changes in one recovery PR.

axpnet:
## Status, criterion by criterion, with the evidence for each

The migration is **implemented on `main`** and **not yet exercised anywhere**. Those are two different things and the checklist conflates them, so here is each box with what actually backs it.

| # | Criterion | State | Evidence |
[...]

PR #485, opened 2026-07-28
ci(snap): gate the whole GPU userspace, not two filename patterns

axpnet:
Closes the last open acceptance criterion of #465: "confirm `snapcraft lint` no longer emits `gpu:` warnings for bundled drivers; assess remaining library-linter warnings individually rather than blanket-ignore them."

## What was already there

`G1` carried an inline grep for `/dri/*_dri.so` and `libEGL/GLX_mesa`. It caught the two loudest patterns and missed the rest of what the provider owns: `libgbm`, the vendor `libdrm` family, vulkan ICDs. It never ran the linter whose `gpu:` warnings opened the issue.

## What this adds

`scripts/snap-gpu-lint-check.sh`, runnable by hand against any `.snap`, wired as **Gate G4** in place of the inline grep:

1. **content pass**, always runs, needs only `unsquashfs`, so it cannot degrade into "could not verify" on a runner;
2. **`snapcraft lint`** when snapcraft is on PATH. `gpu:` warnings fail the gate. `library:` warnings are printed as warnings and do **not** fail it, because the issue asks for them to be judged one at a time (a dynamically loaded library looks exactly like an unused one) rather than blanket-ignored.

## Verified against three snaps, not just read

| Snap | Result |
|---|---|
| synthetic clean | exit 0 |
| synthetic with Mesa inside | exit 1, all three files listed |
| **aeroftp 4.1.6 from the Store** | exit 1, **27 files listed** |

## The third row is the interesting one

It is not a false positive. The published snap has no `graphics-core22` plug and no `command-chain` at all: `snap-refresh` packs with the snapcraft.yaml **of the tag**, and the newest tag predates the migration.

Which means, and this is worth saying out loud on #465: the graphics-core22 migration is on `main` but **has never been packaged or gated end to end**. The gate commit (`d688efadc`, 2026-07-26 23:01) landed *after* the last run that actually packed a snap (2026-07-26 01:56), and the only run since then skipped `pack`. So "CI: G1, G2 and G3 all pass" is **unverified**, not satisfied, and stays that way until a v4.1.7 tag exists.

Note `scripts/snap-gpu-lint-check.sh` needed `git add -f`: `/scripts/*` is gitignored and every tracked script in there was added the same way.

Not merged: opened for review.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved snap validation to detect a broader range of bundled GPU userspace files.
  * Added checks for GPU-related lint warnings before installation and GUI readiness validation.
  * Prevented provider-owned graphics libraries from being shipped unintentionally while allowing approved graphics mount content.

### DIFF
--- a/.github/workflows/snap-refresh.yml
+++ b/.github/workflows/snap-refresh.yml
@@ -266,11 +266,17 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y --no-install-recommends squashfs-tools binutils file
           ./ci-tools/scripts/snap-abi-check.sh "$SNAP_FILE" snap/snapcraft.yaml
-          if unsquashfs -ll "$SNAP_FILE" | grep -E '/dri/[^/]+_dri\.so|/lib(EGL|GLX)_mesa\.so'; then
-            echo "::error::the application snap still bundles Mesa drivers/loaders; graphics-core22 must provide them"
-            exit 1
-          fi
-          echo "OK: GPU userspace is supplied only by graphics-core22."
+
+      # Was an inline grep for `/dri/*_dri.so` and `libEGL/GLX_mesa` inside G1.
+      # It covered the two loudest patterns and missed the rest of what the
+      # provider owns (libgbm, the vendor libdrm family, vulkan ICDs), and it
+      # never ran the linter whose `gpu:` warnings started this migration.
+      # Both now live in a script that can be run by hand against any .snap.
+      - name: Gate G4 - no bundled GPU userspace
+        shell: bash
+        env:
+          SNAP_FILE: ${{ steps.snap-check.outputs.snap }}
+        run: ./ci-tools/scripts/snap-gpu-lint-check.sh "$SNAP_FILE"
 
       - name: Install the graphics userspace provider
         shell: bash
--- a/scripts/snap-gpu-lint-check.sh
+++ b/scripts/snap-gpu-lint-check.sh
@@ -67,12 +67,41 @@ fi
 TMP="$(mktemp -d)"
 trap 'rm -rf "$TMP"' EXIT
 
-# Same defensive form as the ABI gate: a snap carries device nodes and file
-# capabilities, so unsquashfs can report restore errors that are not our
-# problem. Tolerate a non-zero exit and judge on what actually landed.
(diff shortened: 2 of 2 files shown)


axpnet:
> Line 73 swallows all nonzero exits, but `unsquashfs` can return nonzero for non-fatal issues while still writing files, or fatally abort if the snap image is invalid.

Founded, and worse than "can": reproduced. On squashfs-tools 4.6.1, a 301-file image carrying `dri/zzz_iris_dri.so` and corrupted mid-image extracts **182 of 301 files without the driver**, exits 1, and leaves the destination directory in place — so `[ -d "$TMP/snap" ]` passed and the scan ran on a truncated tree. The gate printed `OK: no provider-owned GPU userspace inside the snap.` and exited **0** on a snap that carries one. For a gate whose only job is to fail on presence, that is the worst possible failure mode, and it is the same shape as queued #32 where a retry loop turned green on a Docker Hub outage.

Fixed in `315cc4431`, but not by narrowing the tolerance. `-no-exit-code` would still leave the check reading a tree that may be partial, so the extraction is gone entirely: the check now reads the directory table with `unsquashfs -l` and never unpacks. There is no restore step to fail on device nodes or capabilities — which is the only reason the tolerance existed — and a listing cannot half-succeed. On that same corrupt image `-l` reports all 306 entries with the driver among them, so the gate now fails where it used to pass. A directory table too damaged to read exits non-zero and is a hard error, and a listing of fewer than two entries is refused rather than read as a clean snap, since an empty listing would otherwise sail through as "no offenders".

Two details that came out of the rewrite: `-l` rather than `-ll`, so a path containing a space stays intact instead of having to be cut out of columns; and both `*/dri/*_dri.so` and `dri/*_dri.so` are matched, because paths are relative once `squashfs-root/` is stripped and the first glob alone would have missed a `dri/` directory at the root of the snap, which the previous absolute-path `find` did match.

Five fixtures, all `--no-snapcraft`: corrupt-with-driver 1 (was **0**), intact-with-driver 1, clean-including-an-empty-`graphics/`-mount-point 0, root-level `dri/` 1, destroyed superblock 2.

PR #487, opened 2026-07-28
test(aerorsync): live lane 3 acceptance for the batch and symlink xattr paths

axpnet:
Closes the two acceptance gaps that #484 and #482 each stated they were leaving open. Both are the same shape: the behaviour was true by construction and had never been observed against a real rsync. **Both new tests were run against the defective code they guard**, not merely watched to pass.

## The batch path (R3, follow-up to #484)

#484's own test reaches only the flag inheritance, and says so: the four call sites want a live peer and `AerorsyncBatch` holds a concrete `RusshSessionTransport`, so no mock gets there. Until now "the batch preserves xattrs" was an argument, not an observation.

`delta_upload_xattr_batch_two_files_live_lane_3` uploads two files over one `begin_batch()` session with `.with_xattrs(true)` and verifies, on the remote, both attribute values and both sha256 sums. Deliberate choices:

- **`session_count == 1` is asserted.** Reuse is the point of a batch, so a silent degradation to one handshake per file fails here instead of passing quietly.
- **One inline value and one out-of-band value**, so both xattr encodings cross the batch leg rather than just the cheap one.
- **`!is_noop()` is asserted.** `begin_batch` degrades to `NoopBatch` when russh cannot connect, and `NoopBatch::upload` then fails with a message about session reuse that says nothing about xattrs. The assertion names the real cause.
- A cross-check that neither file carries the other's attribute, since a single shared session is exactly where that could happen.

Worth noting: this is also the **first live exercise of the russh leg** on lane 3. The single-file tests run on libssh2, because `prefers_russh_leg()` is false for a pubkey-file profile, while `begin_batch` always connects through russh.

**Verified as a pin:** with the four call sites put back to the `false` literals of the pre-#484 shape, it fails on `[batch-a] remote xattr user.aeroftp.batch_inline must match the local value`, left `None`.

## The symlink path (R2, follow-up to #482)

#482 pins the local wrappers; nothing pinned the wire consequence. `delta_upload_symlink_xattr_not_inherited_live_lane_3` uploads a link whose *target* carries a `user.*` attribute and asserts the remote link is bare, with two remote probes: `lgetxattr` on the link must find nothing, and `getxattr` *through* the link must find the target's attribute. The control is load-bearing — the existing helper follows symlinks, and a dangling remote link would answer ENOENT and satisfy the assertion for the wrong reason.

**The attribute is deliberately out of band, and that is the difference between a pin and decoration.** With an inline value this test cannot fail on Linux whatever the code does, because the kernel forbids `user.*` on a symlink and the remote link ends up bare either way. That is measured, not assumed: the inline version was run against the reverted wrappers and passed. Above `MAX_FULL_DATUM` the pre-R2 read emits a per-file section for an entry the peer is not expecting, and the stream desynchronises.

That is also a **correction to what #482 predicted**. The failure is not the receiver's EPERM but a hard `HardRejection(InvalidFrame): expected NDX_DONE (0x00), got 0x01` — the same class of desync that the generator's missing `send_xattr_request` drain caused in queued #31.

## CI registration

Both tests are added to the lane 3 filter in `.github/workflows/aerorsync-protocol.yml`. A test missing from that list does not run in CI and nobody notices.

## Gates

| Gate | Result |
|---|---|
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --lib --features aerorsync --all-targets -- -D warnings` | clean |
| same, under `RUSTFLAGS='--cfg ci_lane3'` (lints the new tests) | clean |
| `cargo test --features aerorsync --lib aerorsync` | 597 passed, 0 failed |
| `RUSTFLAGS='--cfg ci_lane3' cargo test --features aerorsync --lib live_lane_3 -- --test-threads=1` | **11 passed, 0 failed** |

Measured against the stock rsync 3.2.7 harness on `127.0.0.1:2224`: batch-a carried an 18-byte inline value, batch-b a 96-byte out-of-band one, both with sha256 parity.

## One practical note for lane 3 from a fresh worktree

`git worktree add` recreates `src/aerorsync/capture/keys/id_ed25519` under the current umask, and ssh refuses a key at 0664. CI does `chmod 600` on it (`aerorsync-protocol.yml:160`); locally you have to do the same, or all nine lane 3 tests fail with `Permission denied (publickey)` **instead of skipping** — the skip guard only covers TCP reachability, not an unusable key.

## Summary by CodeRabbit

- **Tests**
  - Expanded live synchronization coverage for extended attributes during batch uploads.
  - Added verification that attributes remain associated with the correct files across multiple uploads.
  - Added symlink-specific checks to ensure attributes are preserved correctly without inheriting attributes from link targets.
  - Added coverage for both inline and out-of-band extended attribute values.

### DIFF
--- a/.github/workflows/aerorsync-protocol.yml
+++ b/.github/workflows/aerorsync-protocol.yml
@@ -199,6 +199,8 @@ jobs:
             aerorsync::delta_transport_impl::tests::delta_upload_xattr_binary_nul_live_lane_3 \
             aerorsync::delta_transport_impl::tests::delta_upload_xattr_none_with_x_flag_live_lane_3 \
             aerorsync::delta_transport_impl::tests::delta_download_xattr_live_lane_3 \
+            aerorsync::delta_transport_impl::tests::delta_upload_xattr_batch_two_files_live_lane_3 \
+            aerorsync::delta_transport_impl::tests::delta_upload_symlink_xattr_not_inherited_live_lane_3 \
             --nocapture
       - name: Tear down harness
         if: always()
--- a/src-tauri/src/aerorsync/delta_transport_impl.rs
+++ b/src-tauri/src/aerorsync/delta_transport_impl.rs
@@ -3772,33 +3772,62 @@ mod tests {
         let func = if nofollow { "lgetxattr" } else { "getxattr" };
         let cmd = format!(
             r#"python3 - <<'PY'
-import ctypes, base64, sys
+import ctypes, base64, errno, sys
 L = ctypes.CDLL(None, use_errno=True)
 F = L.{func}
 F.argtypes = [ctypes.c_char_p, ctypes.c_char_p, ctypes.c_void_p, ctypes.c_size_t]
 F.restype = ctypes.c_ssize_t
 p = b"{remote_path}"
 n = b"{name}"
+ctypes.set_errno(0)
 s = F(p, n, None, 0)
 if s < 0:
-    print("MISSING")
-    sys.exit(0)
+    e = ctypes.get_errno()
+    if e == errno.ENODATA:
+        print("MISSING")
+        sys.exit(0)
+    sys.stderr.write("probe failed: errno " + str(e) + " " + errno.errorcode.get(e, "?") + "\n")
+    sys.exit(1)
 b = ctypes.create_string_buffer(s)
+ctypes.set_errno(0)
 g = F(p, n, b, s)
-print(base64.b64encode(b.raw[:g]).decode())
+if g < 0:
+    e = ctypes.get_errno()
+    sys.stderr.write("read failed: errno " + str(e) + " " + errno.errorcode.get(e, "?") + "\n")
+    sys.exit(1)
+print("OK " + base64.b64encode(b.raw[:g]).decode())
 PY"#
         );
         let out = lane3_ssh_testuser(key_path, &cmd);
-        if out == "MISSING" || out.is_empty() {
-            None
-        } else {
-            use base64::Engine as _;
-            Some(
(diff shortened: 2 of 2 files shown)


axpnet:
> The probe treats any first-call xattr failure as `MISSING` and also maps zero-length successful output as `None`, so permission/path errors or an empty user attribute can make absence assertions pass without a valid probe.

Founded on all three counts, and the first one is the serious one: it makes every `expect_absent` assertion in this file — including the R2 symlink pin this PR adds — capable of passing without the probe ever having worked.

Reproduced against the lane 3 harness on a path that does not exist:

| | stdout | exit |
|---|---|---|
| old probe | `MISSING` | 0 → `None` → assertion passes |
| new probe | `probe failed: errno 2 ENOENT` (stderr) | 1 → `lane3_ssh_testuser` panics |

Fixed in `36dac6a52`. Only `ENODATA` is absence now; every other errno is written to stderr with its symbolic name and exits non-zero, which the ssh helper turns into a panic quoting stderr.

The other two are fixed in the same pass. The second `getxattr` return was never checked, so a value that shrank between the size call and the read gave a negative `g` and `b.raw[:g]` **silently dropped bytes from the end** rather than failing — a wrong value, not a detected error. And the output is now prefixed (`MISSING` / `OK` / `OK <base64>`) so an attribute whose value is legitimately empty is distinguishable from no output at all; the bare `OK` case is needed because the ssh helper trims trailing whitespace.

Lane 3 re-run after the change: 11 passed, 0 failed. `lgetxattr` on a symlink does answer ENODATA, so the R2 pin still holds — and now it holds for the right reason rather than by accident.

PR #488, opened 2026-07-28
fix(dag): give the durable checkpoint store to the caller instead of reaching for the user's config

axpnet:
## The defect

`execute_single_file_dag` opened `TransferCheckpointStore::default_store()` itself, with no way to override it. The multipart path does not merely *write* there: it scavenges orphan records past the 7-day TTL whose endpoint matches this transfer, and calls `abort_multipart_upload` on each one.

Under `cargo test`, that store is the developer's real `~/.config/aeroftp-dev/transfer-checkpoints`. So every test that shaped a multipart upload read and wrote a real user's configuration, and inherited every previous run's leftovers. On this machine that directory held **325 records accumulated since 21 July**, all from mock endpoints, all written by the test suite.

## Why this matters beyond hygiene: it is the second cause of the #470 flake

`cancel_token_keeps_durable_multipart_session_for_resume` asserts `multipart_aborted == false`. Four `slow-mock` records written on 21 July crossed the TTL on the morning of 28 July; the scavenger aborted one, `SlowMockProvider::abort_multipart_upload` raised the shared flag, and the test failed on exactly the assertion it exists to defend.

The run that fails then `remove()`s the record, which is why the very next run passes. The observed fail/pass pair is the mechanism, not noise.

CI never saw any of it, because a fresh runner has no history. It reproduces only on a developer's machine — the worst place for a flake to live.

**Queued #30 is not retracted by this.** #470 closed a real timing race by construction, and there is indeed no load level at which *that* window returns. This is a different abort path that has nothing to do with when the cancel lands. The tracker row wants an addition, not a correction.

## The fix

The store is a parameter now.

| Caller | Passes |
|---|---|
| `provider_commands.rs` download + upload, `aeroftp_cli.rs` | `None` → the real per-user store, unchanged behaviour |
| 14 call sites in the unit test module | that test's own temp directory |
| 3 call sites in `tests/` (`dag_b2_multipart`, `dag_s3_multipart`, `integration_gtc_wan_segmented`) | same |

Explicit rather than defaulted, on purpose: a new test cannot reach the developer's configuration by forgetting an argument, because it will not compile without one. The three call sites under `tests/` were found by `clippy --all-targets`, not by grep — they were writing to the real store too.

## The pin

`scavenger_aborts_a_stale_record_from_the_injected_store` covers the mechanism that had nothing exercising it, in **both** directions, because "aborts too eagerly" and "never aborts" are both regressions and one assertion cannot tell them apart:

1. seeds a genuine non-terminal record by running a cancelled multipart upload, rather than hand-writing one that could drift from the shape production journals;
2. asserts that record is **not** stale under the default 7-day TTL, so a live resume candidate is left alone;
3. re-runs against the same directory through a zero-TTL view — which is what "seven days later" looks like without waiting seven days — and asserts the session is aborted.

## Gates

| Gate | Result |
|---|---|
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --lib --all-targets -- -D warnings` | clean |
| `cargo test --lib transfer_dag_single_file` | **29 passed, 0 failed**, including the previously flaky test |

## Credit

Root cause independently identified by the triage tab; full evidence in `20260728T074500Z-E4-flaky-multipart-cancel-ROOTCAUSE.md`. The four `slow-mock` records that reproduce it have deliberately **not** been deleted.

### DIFF
--- a/src-tauri/src/bin/aeroftp_cli.rs
+++ b/src-tauri/src/bin/aeroftp_cli.rs
@@ -9022,6 +9022,8 @@ async fn cli_run_single_file_dag(
         file_size,
         // The CLI relies on process-level Ctrl+C, not the GUI session token.
         None,
+        // The real per-user checkpoint store: this is a production transfer.
+        None,
     )
     .await;
     let provider = arc
--- a/src-tauri/src/provider_commands.rs
+++ b/src-tauri/src/provider_commands.rs
@@ -2296,6 +2296,8 @@ async fn run_dag_download_leaf(
         report_size,
         file_size,
         cancel_token,
+        // The real per-user checkpoint store: this is a production transfer.
+        None,
     )
     .await
     {
@@ -2445,6 +2447,8 @@ async fn run_dag_upload_leaf(
         report_size,
         file_size,
         cancel_token,
+        // The real per-user checkpoint store: this is a production transfer.
+        None,
     )
     .await
     {
--- a/src-tauri/src/transfer_dag_single_file.rs
+++ b/src-tauri/src/transfer_dag_single_file.rs
@@ -714,6 +714,23 @@ pub async fn execute_single_file_dag(
     // instead of running the current file to completion. `None` = no cancel
     // wrapping (the CLI path keeps its own Ctrl+C handling).
     cancel_token: Option<CancellationToken>,
+    // Where durable multipart checkpoints are journaled, and therefore which
+    // orphan records this run is allowed to scavenge and abort. `None` means
+    // the real per-user store under the AeroFTP data root, which is what the
+    // three production call sites pass.
+    //
+    // It is a parameter rather than a `default_store()` call inside the body
+    // because the body also *acts* on what it finds there: it aborts stale
+    // provider sessions whose endpoint matches this one. With no way to
+    // inject it, every test that shaped a multipart upload read and wrote the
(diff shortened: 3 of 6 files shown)


axpnet:
Independent confirmation, arrived by accident and worth recording.

Running the full `cargo test --features aerorsync --lib` suite on `fix/russh-0624-ghsa` — a branch that touches `Cargo.toml`, `Cargo.lock` and one russh handler in `aeroftp_cli.rs`, and **nothing** to do with multipart, checkpoints or the DAG — produced:

```
test result: FAILED. 3333 passed; 1 failed; 18 ignored
---- transfer_dag_single_file::tests::cancel_token_keeps_durable_multipart_session_for_resume
panicked at src/transfer_dag_single_file.rs:2456:9:
a durable checkpoint keeps the provider session for missing-part resume
```

One failure in 3334, and it is this one. That branch is cut from `main`, which does not carry this PR's fix.

The store confirms the mechanism rather than merely being consistent with it: `~/.config/aeroftp-dev/transfer-checkpoints` went from **325 files to 319** across that run, so six stale records were aborted and then `remove()`d, with fresh records written at the same timestamp. The machine was running this suite alone — no competing `cargo`, which is the reproducer queued #30 names for the *other* cause.

So the flake is not merely explainable by the scavenger path, it reproduces on demand on an unrelated branch, and it will keep reproducing on every developer machine that has aged records until this PR lands.

PR #496, opened 2026-07-28
feat(blomp): promote Blomp (OpenStack Swift) to production

axpnet:
## What

Blomp moves from dev-only to a **production provider**. It was thought "broken" (403), but the case is solved: the 403 affects **only** the account/container-list operation, while the actual container works fully.

## Root cause (black-box verified against a live account)

- Auth is Keystone v2 with `tenantName=storage` → OK; the catalog exposes the object-store at `swiftproxy.acs.ai.net:8080`.
- `GET {storage_url}?format=json` (list containers = account) → **deterministic HTTP 403**, for every User-Agent, 10/10. This is **by design** (rclone hits the same wall: [rclone#7948](https://github.com/rclone/rclone/pull/7948) is still unmerged).
- There is exactly **one** container per account, fully usable (list/put/get/mv/rm), **named after the login email**.

Our `discover_container()` relied on that account listing, so `connect()` died before touching any file.

## Fix

- **`discover_container()`**: falls back to the container named `= username` (HEAD-verified) on 403/401/empty account.
- **`authenticate()` self-healing**: the version probe was best-effort and a transient failure dropped us to TempAuth v1 → `/auth/v1.0` → 404. Now, if the detected flow fails, the other is attempted before surfacing an error; the probe defaults to Keystone v2.
- **Production promotion**: `stable: true` in the registry, `blomp` removed from `DEV_ONLY_LOGO_IDS`, README + `docs/PROVIDERS.md` regenerated (51 providers).
- **CLI**: `--protocol blomp` is now a known protocol, normalized to the GUI's canonical form (`swift` + `providerId=blomp`) via `canonical_protocol_and_provider()`, so dispatch/capabilities/features (all keyed on "swift") work with no scattered special cases. Removed the now-empty `DEV_ONLY_PROTOCOLS` machinery.

## Live verification (real account)

`connect · ls · df (40 GiB) · upload · download (content match) · rename · delete · 8/8 auth stability` — all green. New unit tests cover the alias normalization. The AeroCrypt encrypted overlay was also verified end-to-end on Blomp (init/put/ls/get, ciphertext-only on the server confirmed independently).

## Local gate

`cargo fmt --check` · `cargo clippy --lib --bin aeroftp-cli -D warnings` · `vitest` · `i18n 46/46` · `cargo test --lib swift` + new bin tests — all green.

## Summary by CodeRabbit

* **New Features**
  * Added Blomp as a supported provider (Swift), including it in the provider catalog and documentation/provider totals.
* **Bug Fixes**
  * Improved Swift/Blomp authentication with automatic auth-version detection and retry-based recovery.
  * Enhanced Blomp container discovery with permission-safe fallback behavior and clearer failure messaging.
  * Updated CLI profile creation to normalize the Blomp protocol alias and store the canonical selection.
* **Documentation**
  * Refreshed provider documentation and provider matrix entries (including updated totals).
* **Tests**
  * Added unit tests covering Blomp protocol normalization for profile-add.


axpnet:
Addressed the CodeRabbit review in `1b468c3b6`:

- **PII in logs (swift.rs)**: valid. The Blomp container name is the login email, so it is now dropped from both the fallback success `info!` and the not-reachable error message.
- **Auth timeouts (swift.rs)**: valid. Added a 30s per-request `AUTH_REQUEST_TIMEOUT` on `detect_auth_version`, TempAuth v1 and Keystone v2, so an accepting-but-stalling endpoint cannot hang connect for the client's long body read timeout.
- **Dispatch comment (aeroftp_cli.rs)**: valid. Reworded to say `profile-add` now persists the canonical `swift` + providerId `blomp` form, so the `"blomp"` arm is legacy-profile compatibility, not a dev-only alias.
- **Persistence-contract test (nitpick)**: skipped, with reason. `persist_new_profile_entry` needs a real (temp) vault, a master password and a `Cli`, i.e. a heavyweight integration harness the surrounding unit tests (all pure-function) do not use. The contract is already verified live (a `--protocol blomp` profile persists `protocol=swift`, `providerId=blomp`) and the call site is a single line consuming the canonicalized values. Not worth a vault-backed integration test for a trivial nit.

Re-verified live after the changes (connect, ls, df 40 GiB) and green: fmt, clippy -D warnings (lib+bin), swift + alias unit tests.

PR #498, opened 2026-07-28
aerorsync: advertise only the compressors we can actually drive

axpnet:
The negotiated compressor is the first name in **our own** advertised list that the peer also offers, computed the same way on both sides (`compat.c::parse_negotiate_str`). That makes the advertisement a promise rather than a wish list: every name in it is one the negotiation may hand back to us.

We were advertising stock rsync's list verbatim, `zstd lz4 zlibx zlib`, while this module drives exactly two of those four.

## The defect

It is worse than a harmless superset, because stock ranks `lz4` above `zlibx`. A peer built with lz4 but without zstd made the negotiation settle on a codec that has no implementation here at all, so the transfer fell back to the classic wrapper while a working, byte-pinned `zlibx` sat one position lower. We were losing to our own advertisement.

Same shape as the bug #491 just closed: a declaration that promised more than the code delivers.

## The fix

`zstd zlibx none`. The trailing `none` is the load-bearing part: it guarantees the two lists always intersect, so a peer offering only codecs we decline lands on uncompressed literals, which is a working delta transfer, instead of an empty intersection. On Windows that difference is the whole native path, since there is no classic fallback there.

## Why plain `zlib` stays out, structurally rather than for effort

`zlib` feeds matched block data through the compressor history on both ends (`token.c::see_deflate_token`), so decoding it needs `inflateIncomp`, a function rsync patches into the zlib it bundles. It is in neither the system zlib (checked with `nm` on the `.so`) nor any Rust crate. Upstream split `zlibx` out for exactly this reason, and the rsync man page says so directly:

> the "zlibx" compression algorithm is just the "zlib" algorithm with matched data excluded from the compression stream (to try to make it more compatible with an external zlib implementation)

Supporting it would mean owning an inflate implementation. The correct move is never to negotiate it, which is now a property rather than a hope.

## Tests

Four new, and **verified as pins by restoring the old list and watching them fail**, not by watching them pass:

| Test | What it catches |
|---|---|
| `lz4_capable_peer_without_zstd_still_lands_on_zlibx` | the defect itself |
| `zlib_only_peer_lands_on_uncompressed_literals_not_a_typed_error` | the working-delta guarantee the trailing `none` buys |
| `default_advertisement_never_yields_a_codec_we_cannot_drive` | the rule, not an example: a sweep over all 32 subsets of stock's own compress list, so a name added without a codec fails here instead of against a stranger's NAS |
| `forcing_zlib_through_the_env_shaped_profile_still_takes_the_typed_error` | keeps the `Unsupported` arm covered, since making the default safe would otherwise silently delete the coverage the classic fallback depends on |

Three byte-pinned driver tests re-declared the advertisement by hand, which is the duplication that let the string and its pins drift apart. They now derive it from `PreambleProfile::default()` and check that the driver **emits what it advertises**, leaving `preamble_profile_default_is_byte_pinned` as the single place pinning the literal strings. Confirmed the split works: with the defect restored, the pin test fails and the three derived ones pass, which is each doing its own job.

## Also

A stale number that predates this change: the `real_wire` module doc called the client preamble "55 bytes", which was an observed peer shape and never ours. With the shipping profile we emit 46, and the derivation is written out so the next reader can check it rather than trust it.

Docs realigned in all three places, including the `lz4` row that did not exist anywhere.

## Gate

605 to 609 aerorsync tests, 0 failed. `cargo fmt` clean, `clippy --lib --features aerorsync --all-targets -D warnings` clean, `RUSTFLAGS='--cfg ci_lane3' check` clean.

## Summary by CodeRabbit

* **Compatibility**
  * Improved compression negotiation by advertising only `zstd`, `zlibx`, and `none`, and selecting the first codec both sides offer.
  * Added/strengthened fallback to uncompressed literals when peers only support unsupported codecs (e.g., `zlib`).
  * Updated default negotiation behavior to avoid advertising codecs that can’t be safely handled.
* **Documentation**
  * Clarified compressor negotiation and the “winner” selection logic across the README and protocol comparison docs.
* **Tests**
  * Expanded coverage to verify negotiation outcomes across different peer codec sets.


axpnet:
Nitpick accepted and closed in `11c1777f2`, and it was worth more than a nitpick: the blind spot it names is the same species of defect this PR exists to fix, reproduced inside the fix itself. The sweep was anchored to a `STOCK` list sitting next to the test rather than to the thing it guards, so a codec added to `PreambleProfile::default` that stock rsync never heard of would never be offered by the synthetic peer and the test would have stayed green while we advertised something undrivable.

Implemented as suggested, with the assertion driven from `PreambleProfile::default()` itself. Verified as a pin rather than assumed: with a fake `brotli` inserted into the default advertisement the old test passes and the new one fails with

```
the default advertisement carries "brotli", which nothing here can drive:
Unsupported("brotli"). Either add the codec or stop advertising the name
```

The two sweeps now cover each other: the subset one for combinations stock can produce, the new one for anything we ship that stock does not list.

PR #499, opened 2026-07-29
chore(ci): watch the n0-mainline patch instead of writing down that someone should

axpnet:
`src-tauri/Cargo.toml` carries a `[patch.crates-io]` pin on a one-line fork of n0-mainline, added in #489 so russh 0.62.4 — which closes GHSA-g9hv-x236-4qp3, GHSA-5xvq-cp9x-6p6r and GHSA-cqjc-rmpq-xprq — could resolve against n0-mainline's `=3.0.0-rc.0` pin on ed25519-dalek. The section has always carried an exit condition in a comment: delete it once the graph resolves without it. A comment is not a mechanism, and a `[patch.crates-io]` nobody revisits is how a project keeps shipping a fork it no longer needs and no longer audits.

## The check that had quietly become a liar

`.github/scripts/check-russh-unblock.sh` asked *does russh 0.62.4 resolve?*. That was the right question until #489 answered it: it resolves, **because of the patch**. The script reads the requirement from the manifest, and the manifest now says `0.62.4`, so its scratch edit is a no-op and the resolution it tests is the one already committed. Run today it prints `UNBLOCKED - russh 0.62.4 now resolves`, tells the reader to bump russh to 0.62.4 and close known issue #2 — both already done — and exits 0. Verified before removing it rather than assumed.

A check whose subject has been fixed does not become harmless, it becomes a false green. This one would have been read as "the fork can go" by whoever ran it next.

## What replaces it

`check-n0-mainline-patch.sh` asks the question that is actually still open — **is the patch section still needed?** — and answers it by doing the thing rather than by asking upstream about it: strip the section in a scratch copy of the manifest, `cargo generate-lockfile` to re-resolve the whole graph from scratch (the committed lock still points at the patched git source, so reusing it would prove nothing), and restore `Cargo.toml`/`Cargo.lock` unconditionally on the way out. It refuses to run against a dirty manifest, since it rewrites both files.

Testing resolution directly is what makes it correct **however** the block clears — an upstream release, a change in the dependency chain, a russh bump — with no knowledge of GitHub, of n0-mainline's release cadence, or of any API. It also self-terminates: with the section gone it exits 0 saying there is nothing left to watch.

## Both branches verified, not just the one that fires today

| Tree under test | Result |
|---|---|
| as committed | `STILL NEEDED`, **exit 1**, reproducing the exact `ed25519-dalek` conflict |
| russh put back to `0.61.2` — what the graph looked like before #489, so a real instance of "the conflict is gone" | `NOT NEEDED`, **exit 0**, through a genuine cargo resolution |

The exit-0 report prints the resolved `n0-mainline` version on purpose: a **downgrade** also makes the conflict disappear, and that is not the outcome this is waiting for, so the distinction goes in front of the reader instead of hiding behind an exit code.

## The workflow

Mondays 05:23 UTC plus `workflow_dispatch`. It stays **silent** on exit 1, which is the normal state and expected to last months — a watcher that reports "still blocked" every week trains its reader to ignore it. On exit 0 it files a single chore issue, guarded by an exact-title search so the weekly schedule cannot open one per week. On exit 2 (the failure changed shape, or the check itself broke) it **fails the job**, because a watcher that cannot tell "still blocked" from "I no longer work" is worse than none. No `rust-cache`: a restored cache is exactly the thing that could answer with yesterday's index.

## Scope

No dependency and no Rust source is touched. The `Cargo.toml` diff is **comment lines only** (verified mechanically), and the security evidence doc gains a resolution section plus how the remaining item is tracked, replacing its reference to the deleted script.

Local validation: `bash -n`, YAML parse, the awk strip checked against the real manifest with `tomllib` (section and entry removed, TOML still valid, `patch` key absent), and the two script runs above. The heavy cargo gate was **not** run: no `.rs` changed, and another tab was building at the time — flagged rather than skipped silently.

## Summary by CodeRabbit

* **Chores**
  * Added a weekly automated check to determine whether the `n0-mainline` dependency patch is still required.
  * Automatically reports when the patch can be removed, while avoiding duplicate tracking issues.
  * Removed the obsolete `russh` unblock helper script.

* **Documentation**
  * Updated security evidence and dependency guidance with the current resolution and monitoring process.
  * Clarified the conditions for removing the temporary patch.


axpnet:
Closing the loop on the review here: I merged this before reading the comments, and all three were right. Fixed in #503.

> Any failure that mentions `ed25519-dalek` — including a download or registry failure — is treated as the expected block and exits 1.

This was the real one. Exit 1 is the state the workflow deliberately stays silent on, so a permanently broken check would have reported nothing every Monday — the exact confusion this PR's own description said the exit-2 path existed to prevent. Now matched on the resolver-conflict signature (`failed to select a version for` + the crate + `previously selected package`), with no version literals so an upstream move to another release candidate still matches. Verified against a canned transport error that names the crate twice: it now exits 2 where the old predicate exited 1.

> `git diff --quiet` only compares the worktree with the index. A staged `Cargo.toml`/`Cargo.lock` change passes this guard.

Correct, and worse than described: the restore was `git checkout -- <path>`, which restores the **staged** content, so the script could have run against staged work and later reinstated it. Both sides fixed — the guard checks `--cached`, and the restore is `git checkout HEAD --`.

> If the chore is closed before the patch section is removed, the next weekly run creates a duplicate issue.

Right. Now `--state all`, and a closed-but-still-true chore is reopened with a comment rather than duplicated.

The `Cargo.toml` downgrade note was also added where someone deleting the section would read it.

PR #505, opened 2026-07-29
test(linux): prove the file chooser leaves the process, and find out what happens when it cannot

### LINKED ISSUE #464: test(linux): add deterministic portal-backed file chooser smoke coverage
## Context

PR #463 adds the blocking Snap G3 startup/rendering gate for #462. During review, CodeRabbit correctly noted that startup readiness does not exercise AeroFTP's file/folder chooser path.

On Linux, `src-tauri/src/lib.rs` deliberately defaults `GTK_USE_PORTAL=1` because the in-process GTK directory chooser has previously triggered GLib heap corruption under WebKitGTK. This is a distinct user flow from the blank-window/EGL startup regression fixed by #463.

## Required design

Add deterministic Linux GUI coverage that:

- invokes a real AeroFTP file or folder picker after `app_ready`;
- proves whether the request is portal-backed when a portal service is available;
- closes/cancels the chooser automatically without human interaction;
- distinguishes a clean user cancellation from portal rejection, crash, or hang;
- runs under an isolated X/DBus/portal session and never opens on the owner's desktop;
- does not expose a production automation/backdoor command in normal builds;
- retains the existing heap-corruption, EGL, splash-timeout and visibility assertions.

A test-only feature/build or a separately compiled harness is preferable to a runtime environment-variable backdoor in the release binary.

## Acceptance criteria

- [ ] Deterministic chooser open + cancel under CI.
- [ ] Explicit positive proof of portal handling where portal is configured.
- [ ] Negative test for rejected/unavailable portal with documented GTK fallback behavior.
- [ ] No human desktop interaction.
- [ ] No production Store publication until its owning release gate passes.

Origin: CodeRabbit discussion in #463 (`discussion_r3651982704`).

axpnet:
Closes #464.

Deterministic, headless coverage that AeroFTP's file chooser really leaves the process through `xdg-desktop-portal` — and, it turns out, what happens when it cannot.

## Why this is worth a gate

`src-tauri/src/lib.rs` forces `GTK_USE_PORTAL=1` because the in-process `GtkFileChooser` has corrupted the GLib heap under WebKitGTK. That makes "the chooser runs out of process" a **safety property**, not a preference — and nothing inside the app can prove it holds. Gate G3 proves the window paints; it never opens a chooser.

Verified rather than assumed, since the whole issue rests on it: `rfd 0.16`, which `tauri-plugin-dialog` uses, resolves against **`gtk-sys`** in this tree and `ashpd` is absent, so the chooser really does go GTK → portal.

## How it drives the app

Three named controls — **Export / Import → Import Servers → Choose .aeroftp file...** — each activated through AT-SPI. **Not clicked.** A coordinate that misses is indistinguishable from a control that did nothing, so a click-driven version would pass while proving nothing; a name that is absent exits 4 and a control with no action exits 5, and neither is ever success.

No test-only feature, no build flag, nothing added to the shipped binary. The stand-in is a standalone crate; CI asserts `src-tauri` never references it.

**The evidence that the call is real is the handle token.** GTK generates `gtk<random>`, so a recorded `"handle_token":"gtk573679125"` beside the app's own dialog title (`Import Servers`) says the request came through the toolkit rather than through anything the harness staged.

| Case | Asserts |
|---|---|
| portal cancels | the portal **is** called, `OpenFile`, GTK token, app's title, answered cancelled, app healthy |
| portal refuses | still called, refusal delivered as a D-Bus error, no panic, no hang |
| portal not installed | pins the behaviour **measured** (below) |

**Verified as pins, not by watching them pass:** pointed at controls that open no chooser, 7 assertions fail and the gate exits 1. The stand-in has its own 9-assertion self-test, likewise proven by breaking the Request path.

## What it found

`lib.rs` says that on a host with no portal, GTK falls back to the native chooser. **It does not.** Measured on a bus that activates all 65 other session services and omits only the 9 portal ones — a faithful model of a machine without `xdg-desktop-portal` — no chooser window appears and nothing is presented to the user. Window count after the press is identical to the portal-backed run.

The frontend does not soften it: [`ExportImportDialog.tsx`](src/components/ExportImportDialog.tsx#L172) calls `open()` with no `try`/`catch`, so a refusal becomes an unhandled rejection and the user sees no message. On a minimal WM, in a container, or on any install without the portal, **the picker silently does nothing**.

The gate pins that as measured rather than as wished, with a comment saying so, so a future fix has to change the assertion deliberately. **The fix is not in this PR** — it is an app behaviour change and deserves its own.

## Isolation: three real leaks, now asserted rather than assumed

- **Wayland defeats `DISPLAY`.** The session carries `WAYLAND_DISPLAY` and `GDK_BACKEND=wayland`, and GTK prefers Wayland — so pointing `DISPLAY` at a private Xvfb isolated nothing: launches opened on the live desktop while the harness captured a blank frame from an Xvfb nobody was drawing on. That blank frame was the evidence, misread as a capture artefact. Now `WAYLAND_DISPLAY` is unset, `GDK_BACKEND=x11` is forced, and the run **fails** unless a window with class `aeroftp` exists on the private display.
- **`dbus-run-session` does not isolate the accessibility bus.** Separate bus, own fallback chain, which resolved to the real desktop — `gnome-shell`, the browser and the editor were enumerable. Since the trigger activates by name, that was a route into someone else's application. A private bus is started explicitly and asserted before anything is pressed. `XDG_RUNTIME_DIR` is deliberately **not** overridden: that breaks GTK init outright, and the bus is per-display anyway.
- **`dbus-run-session` activates the real portal on demand.** With the stand-in absent, the system portal started by activation and answered in its place, turning "no portal installed" into "the real portal on a private bus".

## CI

`selftest` (D-Bus only, under a minute) blocks every PR touching the harness. `gate` builds the app and drives it; it runs on the paths that can break it, on demand, and weekly — a full Rust build is not worth spending on unrelated PRs. Evidence is uploaded as an artifact either way.

Acceptance criteria 1–4 are met. Criterion 5 (no Store publication until the owning release gate passes) is a release-time hold, unaffected by this.

## Summary by CodeRabbit

* **Tests**
  * Added a Linux portal-backed integration test suite for file/folder chooser flows (cancel, accept with returned URIs, portal refusal, and behavior when no portal is available).
  * Introduced a deterministic portal stand-in and self-check to ensure it’s only used in tests.
  * Added accessibility bus “privacy” verification plus optional AT-SPI inspection/control utilities, with evidence capture from runs.
  * Expanded checks to include portal-backed network monitoring/proxy resolution.
* **Documentation**
  * Added README instructions describing the harness layout, expected outcomes, and how to run both self-tests and full gate runs.


axpnet:
## CodeRabbit round: all 17 addressed, none dismissed

Every comment was checked against the code before acting, and all 17 turned out to be real. Pushed in 38fe67633.

Most of them are one defect in different clothes — **a check that cannot fail, or one that treats absence as proof**. That matters more in this harness than in ordinary code, because the only thing it sells is trustworthy evidence.

### The four that were load-bearing

| Finding | Why it mattered |
|---|---|
| Wayland guard is tautological | It tested `WAYLAND_DISPLAY`/`GDK_BACKEND` right after assigning both, so the check its own comment calls "the single most important thing in this file" could never fire. It now records what was **inherited** and refuses on a reachable compositor socket. Confirmed on a Wayland desktop: exits 2 before Xvfb starts, where the old one sailed through. |
| Empty a11y bus passes | The gate greps the success line as its proof of isolation, so an unloaded atk-bridge read as "the bus is private". Now refuses when nothing of ours is on the bus, and says why. |
| `sed` rewrite unverified | If `private-bus.conf` ever stops carrying that literal element, the substitution is a silent no-op, the system service dirs survive, and the real `xdg-desktop-portal` is activated on demand — turning "no portal installed" into "the real portal on a private bus", which is the exact substitution case 3 exists to prevent. Now verified, fails closed. |
| `cleanup` trap reaps nothing | `APP_PID`/`PORTAL_PID` are assigned inside the `bash -c` child, so both lines were no-ops in the shell that ran them and an interrupted run orphaned the app and the stand-in. PIDs are now recorded under `$OUT` and reaped from there. |

### `Request.Close()` reached nothing, and nothing noticed

`Request` was exported at the portal path, but callers call `Close()` on the per-call handle they were handed. zbus dispatches on an exact path, so the call landed nowhere and the only symptom was a bus error in the app log that reads like a portal failure.

Fixed by exporting each `Request` on its own handle — and **pinned**, not assumed: the probe now calls `Close()` the way GTK does, as self-test case 5. Demonstrated against the defect rather than watched passing:

```
probe: CLOSE FAILED on /org/freedesktop/portal/desktop/request/1_1/tok_close:
       org.freedesktop.DBus.Error.UnknownObject
passed: 9   failed: 1
```

Restoring the old export path fails that assertion and **nothing else** — all nine others stay green, which is precisely why the omission was invisible.

### Also fixed

The `last_sender` race (zbus runs a task per method call, so two overlapping chooser calls could make one predict its Request path from the other's sender and answer where its own caller was not listening); the duplicated `sanitize_path_element`/request-path formula, now one shared module; swallowed log-write failures, which left an empty evidence file the harness reads as "the portal was never asked"; the `grep -c` double-print; the `set -e` abort in the self-test; the preflight, which now checks what is actually used and no longer reports a missing `xdotool` as "the app opened on the real session"; and the config/README comments that stated the opposite of what the code did.

Two things found while reading, beyond the review: `private-bus-noactivation.conf` was referenced by nothing and is gone, and the **"GTK falls back in-process on refusal"** claim is removed from all four places that stated it as fact. The no-portal case measured the opposite, and the `error` case does not measure it at all — so it is no longer asserted anywhere it has not been established.

### Still open: the gate is red for an environment reason

The runner's webview holds an empty document and `app_ready` never arrives. `WEBKIT_DISABLE_SANDBOX_THIS_IS_DANGEROUS` is a **declared candidate, not a verified fix** — it does not reproduce locally, where the gate is green.

So the failure path now measures instead of inferring: it fetches the app's own frontend over loopback, and the next run says which half is broken. A `200` means the bundle is fine and WebKit did not render it; a failure means the frontend was never served, and no WebKit flag will help.

Self-test 10/10 locally, `cargo fmt --check` and `clippy -D warnings` clean.

axpnet:
## Socket alert on `cargo/tokio@1.53.1`: false positive, and it cannot reach the product

> **Obfuscated code**: cargo `tokio` is 90.0% likely obfuscated — From: `tests/portal-chooser/fake-portal/Cargo.lock`

Three facts settle this, all checked rather than assumed.

**It is a heuristic misfire on a first-party crate.** `tokio` is the standard Rust async runtime, published by the tokio-rs project from crates.io. Socket's own scores for it in this same scan are Vulnerability **100**, Maintenance **100**, License **100** — the only thing firing is the obfuscation heuristic, which trips on tokio's large generated and macro-heavy sources. There is no advisory behind it.

**It is not a new dependency for this project.** The shipped app already builds `tokio` (`src-tauri/Cargo.lock` → `1.52.3`). The harness resolves `1.53.1` in its own lockfile, so Socket sees "added" against a lockfile it had not scanned before, not a new supply-chain surface for AeroFTP.

**It cannot reach a shipped binary.** `tests/portal-chooser/fake-portal` is deliberately not a workspace member and is not referenced by `src-tauri` — that is the constraint #464 was written under: no test-only feature and no automation backdoor in the production app. It is not a claim on trust either, there is a CI step in this same workflow that fails the build if it is ever wired in:

```
if grep -rn "portal-chooser\|aeroftp-fake-portal" src-tauri/Cargo.toml src-tauri/Cargo.lock; then
  echo "::error::the portal harness has been wired into the app manifest; it must stay standalone" >&2
  exit 1
fi
```

### What I am not doing about it

Dropping `tokio` would mean swapping zbus onto `async-io`/smol — another runtime, another set of added packages, no security gain. Pinning the harness back to `1.52.3` to shrink the diff would be contorting the dependency graph to quiet a heuristic, and would leave the test crate on an older runtime than the app it tests. Neither is a fix; both are the alert wearing a different shape.

Marking it acceptable **for this pull request only** in the next comment — not `ignore-all`, and no change to the org triage state, so the alert fires again if this package shows up anywhere it matters.

axpnet:
@SocketSecurity ignore cargo/tokio@1.53.1

axpnet:
## Gate green, sandboxed: 17/17 — and the red was never WebKit

`File chooser is portal-backed` **passes**, with the WebKit overrides removed, so the app under test runs the way users run it. `Portal stand-in self-test` passes at 13/13. The four CodeRabbit findings from this round are all addressed.

### What the failure actually was

The declared candidate `WEBKIT_DISABLE_SANDBOX_THIS_IS_DANGEROUS` was aimed at a misdiagnosis. **The harness was breaking the app it measures.** The stand-in owned `org.freedesktop.portal.Desktop` while answering `UnknownInterface` for everything except `FileChooser` — and owning that name is a promise to the whole desktop stack, not just to the interfaces a test cares about.

The evidence came from one instrumented run rather than from more hypotheses:

| Case | Portal | Result |
|---|---|---|
| `cancel` | present | no `app_ready`, splash safety timeout at 10s |
| `error` | present | no `app_ready`, splash safety timeout at 10s |
| `noportal` | **absent** | `app_ready`, 2 windows, frontend rendered |

The no-portal case producing a full accessible tree — `[document text] 'AeroFTP - Modern FTP Client'` — in the **same job** is what settled it: WebKit renders fine on that runner. The variable was never the renderer.

Under `GTK_USE_PORTAL=1`, GIO routes `g_network_monitor_get_default()` and `g_proxy_resolver_get_default()` through the portal instead of netlink, and WebKit consults both before loading a URL. So it declined `http://127.0.0.1:14321` while that URL was serving a 200 with the real `index.html` — which the harness proved by fetching it on the failure path instead of leaving it to inference.

### What is different now, beyond the fix

`NetworkMonitor` (with `GetStatus`, honouring the advertised v3) and `ProxyResolver` are implemented and pinned. But the durable part is that the harness can no longer hide this class of defect from itself: **a stand-in cannot report the interfaces it does not implement**, so `dbus-monitor` records the whole private bus and the failure path summarises which portal interfaces were asked for and which errors came back. The next missing interface names itself instead of costing a round.

The rule is written into the README: *implement what owning the name promises, not only what the test asserts.*

### On the sandbox

> A release-relevant check can pass in conditions that do not reflect normal sandboxed app behavior.

Agreed, and it is now moot rather than waived: with the real cause fixed the gate is green **with the sandbox on**, and the run above proves it. The flags are removed, not kept as harmless — a check passing only with the web-process sandbox disabled is checking a different program.

### Still open, deliberately not fixed here

The no-portal case pins **measured** behaviour that contradicts `src-tauri/src/lib.rs`: the comment says GTK falls back to the native chooser without a portal, and it does not — no window appears and nothing reaches the user, with `ExportImportDialog.tsx` calling `open()` without `try`/`catch`, so the refusal is an unhandled rejection with no message. That is an app behaviour change and deserves its own PR, not a quiet ride on a test one.

Remaining checks: the four `build-and-release` jobs are still running; everything else is green.

Issue #514, opened 2026-07-29
[BUG] AeroSync > Templates don't export


EhudKirsh:
**Describe the bug**
AeroSync > Templates don't allow exporting in any of the 4 file formats they offer.

**Steps To Reproduce**
1. Go to AeroSync.
2. Go to Templates.
3. Pick any format.
4. Click 'Export script'.
5. Click 'Save'.
6. It returns `⚠ Profile 'String' not found`. I'm not sure if this string is sensitive, so I don't spell it out here. It doesn't save any new file.

**Expected behavior**
Be able to save custom settings of AeroSync to a new template file.

**Environment:**
 - OS: Windows 10
 - AeroFTP Version: v4.1.6

axpnet:
Fixed on `main` (PR #516): every AeroSync template export was looking up the connected saved-server id in the sync-preset store, so all four formats failed with `Profile '...' not found`. The dialog now picks a preset and passes the server name separately.

Detail and commits: https://github.com/axpdev-lab/aeroftp/issues/458#issuecomment-5125260493

Ships in v4.1.7 (not in the v4.1.6 build you tested). Happy for you to retest from `main` or from that release and close if export works on your side.

EhudKirsh:
Thanks for fixing exporting templates in v4.1.7. It seems to work on all four available options.

But importing templates doesn't seem to do anything. I hoped that it would change the settings in the Plan and Sync tabs to what the templates save, to save the user time on picking these settings each time.

axpnet:
Fixed on `main` for v4.1.8: applying an AeroSync template now updates the mounted form, including paths, direction, excludes, parallelism, compression, verification, dry-run, and conflict policy. [Details and verification](https://github.com/axpdev-lab/aeroftp/issues/575#issuecomment-5402719725).

EhudKirsh:
Thanks, I see an improvement here in v4.1.8. Exporting a template set to Mirror mode, switching to another mode and then importing and applying the template switches back to Mirror as intended.

However, the rest of the settings like VERIFY POLICY, COMPRESSION and CANARY MODE are all reset to AeroSync's defaults instead of applying what they were when the template was exported. I checked, and the templates don't keep compression_mode as Auto, which is what I set them to, but always export it as "off". compare_size and compare_timestamp always export as true, and compare_checksum as false, no matter what I choose. CANARY MODE isn't an option in these template files.

Maybe an easier design would be to just keep the user preferences from last time each time AeroSync is opened?
[...]

Issue #511, opened 2026-07-29
[BUG] Can't drag the GUI window from the right of the top bar


EhudKirsh:
**Describe the bug**
Just to the left of AeroVault at the top bar, when I hold that empty-looking space and try to drag it around, the window doesn't move. But it's possible to drag the window from just to the right of 'Add friend' of AeroShare. There might be an invisible element just to the left of AeroVault that is causing this. It's not just dragging, but also double-clicking, which maximises and restores at the left but not the right.

**Steps To Reproduce**
1. You can try this in both a Cyber theme and another theme.
2. Double-click and try to hold and drag the window from the right of the top bar, just to the left of Security Tools or AeroVault.
3. Double-click and manage to hold and drag the window from the left of the top bar, just to the right of AeroShare.

**Expected behavior**
To be able to hold any empty space at the top bar and drag it to move the GUI window around.

**Environment:**
 - OS: Windows 10
 - AeroFTP Version: v4.1.5 & v4.1.6

**Additional context**
I’m guessing this is not an international clue for the Cyber theme Security Tools Easter egg.

EhudKirsh:
I managed to drag the GUI window and restore and maximise it by double-clicking on the right of the top bar, just to the left of AeroVault in v4.1.7. I also tried it in the Cyber theme, just to the left of Security Tools.

PR #516, opened 2026-07-29
fix(ui): restore titlebar dragging on the right, and unbreak AeroSync template export

axpnet:
Three reports from Ehud Kirsh, one commit: two window-level defects in the titlebar and the AeroSync template export, plus the path-bar chevron from the wishlist.

## The titlebar could not be dragged from its right half (#511)

The outer titlebar container deliberately carries no `data-tauri-drag-region`: on Windows the runtime intercepts the mousedown geometrically over a drag region before the DOM event is dispatched, which used to swallow clicks on modal close buttons falling inside the bar's 36px. The affordance was left to two explicit children, and everything to the right of the AeroShare button was neither.

Two dead zones, both now filled with drag regions that contain no interactive children, so the Windows behaviour above cannot come back:

- the reserved 96px navigation slot. It exists so the utility cluster does not shift when Connect and Disconnect swap, and on My Servers and Add Service it renders no button at all: 96 inert pixels immediately left of AeroVault, which is exactly the spot in the report.
- the two 12px inter-cluster gaps, previously a `gap-3` on the flex container. A gap is dead space the window manager cannot see, so they are now explicit spacers of the same width.

This also restores double-click to maximise and restore on that side, which follows the drag region rather than a handler of ours.

## AeroSync template export failed in all four formats (#514)

Export returned `Profile '<id>' not found` for `.aeroftp-script`, `.aerosync`, bash and PowerShell alike. Two objects in this codebase are called a profile, and the dialog was handed the wrong one: the connected **saved server**, while all three export commands resolve the id against the **sync preset** store (Mirror, Two-way, Backup, Pull, Remote Backup). The lookup could never match, so no format could ever have worked.

The dialog now has a preset picker, which is the object the export actually describes (direction, comparison keys, retry policy), and the saved server travels separately as what the generated script connects with.

That second half was a defect on its own, latent behind the first. The export wrote `CONNECT --profile` with the **template name the user typed** into the dialog, and the bash and PowerShell wrappers put the same value in `aeroftp-cli sync --profile`. Since the CLI resolves that name against the vault, even a script that had exported successfully would not have connected. The string in the error message is the saved-server id, not a credential.

## The path-bar chevron now points down while its menu is open (discussion #347)

Pure CSS, `rotate-90` on the open state, matching what File Explorer does. One component covers every surface Ehud named: `BreadcrumbBar` is shared by AeroFile, the local panel and the remote panel. The separator chevrons also gained the `aria-haspopup` and `aria-expanded` the trailing one already had.

## Verification

`tsc` (TS 7.0.2) clean, `npm run build` ok, `vitest` 531/531, `cargo clippy --all-features --tests` exit 0, `i18n:validate` 46/46 clean with 0 placeholders. No new i18n key: the preset picker reuses `syncPresets.title`, which is already translated in all 47 locales, so this PR does not touch the locale files and cannot conflict with the i18n chain.

Frontend only. Branched off `main` at `64efc8f5`, after TypeScript 7.

## Merge order

This is outside the agreed `#502 -> #506 -> #509` chain and blocks nobody, but it shares one file with #506 (`src/App.tsx`, one added line in a different block), so it should land after it.

Reported in #511 and #514, and in discussion #347.

Not marked as closing either issue: they stay open until Ehud confirms on Windows 10.

## Summary by CodeRabbit

* **New Features**
  * Added sync preset selection when exporting synchronization templates and scripts.
  * Exports now use the selected preset and active server profile name for connection details.
  * Export actions remain unavailable until a preset is selected.

* **Bug Fixes**
  * Improved breadcrumb dropdown indicators with clearer open states, rotation, and accessibility attributes.
  * Improved custom titlebar dragging between control areas for more reliable window movement.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10766,6 +10766,7 @@ const App: React.FC = () => {
           initialSource: currentLocalPath || '',
           initialDestination: currentRemotePath || '',
           activeProfileId: activeUnifiedRemoteProfile?.id,
+          activeProfileName: activeUnifiedRemoteProfile?.name,
           isProvider: isProviderConn,
           excludePatterns: [],
           protocol: activeUnifiedRemoteProfile?.protocol,
--- a/src/components/AeroSync/AeroSyncDialog.tsx
+++ b/src/components/AeroSync/AeroSyncDialog.tsx
@@ -91,7 +91,7 @@ export const AeroSyncDialog: React.FC<AeroSyncDialogProps> = ({
         && !!context.activeProfileId;
     const localPath = context.initialSource || '';
     const remotePath = context.initialDestination || '';
-    const profileId = context.activeProfileId || '';
+    const serverProfileName = context.activeProfileName || '';
     const excludePatterns = context.excludePatterns || [];
     const isProvider = !!context.isProvider;
 
@@ -278,7 +278,7 @@ export const AeroSyncDialog: React.FC<AeroSyncDialogProps> = ({
                         onClose={() => setShowTemplates(false)}
                         localPath={localPath}
                         remotePath={remotePath}
-                        profileId={profileId}
+                        serverProfileName={serverProfileName}
                         excludePatterns={excludePatterns}
                     />
                     <MultiPathEditor
--- a/src/components/AeroSync/types.ts
+++ b/src/components/AeroSync/types.ts
@@ -94,6 +94,12 @@ export interface AeroSyncContext {
      * stay hidden.
      */
     activeProfileId?: string;
+    /**
+     * Display name of the same saved server. The Templates dialog writes it
+     * into the exported script's `CONNECT --profile`, which is resolved by
+     * name against the vault, so an id would not do (#514).
+     */
+    activeProfileName?: string;
     isProvider?: boolean;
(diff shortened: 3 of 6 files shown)


axpnet:
All three findings applied in `644e2779`. The Major one was worth catching: a stale preset id surviving a failed reload would have reached the backend and produced the same `Profile not found` this PR exists to remove, so the failure mode was a reappearance of the bug rather than a new one. Cleared before the load instead of only on success, which also let the id-preserving branch go.

On the ARIA popup semantics: correct, and the inconsistency was not introduced here. The trailing chevron already declared `aria-haspopup="menu"` over the same role-less dropdown, and I had propagated it to the separator. Rather than leave two siblings disagreeing, I removed the declaration from both. Implementing real `menu`/`menuitem` semantics would mean arrow-key navigation and focus management this dropdown does not have, and declaring the role without them promises keyboard behaviour that is not there. `aria-expanded` stays, which is accurate on a disclosure button by itself.

The select now has a `<label htmlFor>` tied to it by id.

Re-verified after the change: `tsc` clean, `vitest` 531/531, `i18n:validate` 46/46 with 0 placeholders, `vite build` ok.

PR #519, opened 2026-07-30
feat: Ehud wishlist 2026-07-30 (#347) — AeroSync tab state, duplicate hashes, shared trash table

axpnet:
Ten wishlist comments @EhudKirsh left on 29 and 30 July in #347, plus three follow-ups he had left on threads that were already answered. Eight commits, one per theme.

Three of them were not the polish they were filed as. One is a bug that made an entire feature unusable, one is a number the UI has been printing that was never what it claimed to be, and one is a control that has never reached the code it configures.

## AeroSync deleted the settings you were about to schedule (17832239)

> I'm not sure how I'm supposed to be able to schedule a Plan with Update and other settings if they're deleted when switching to the Sync tab.

He is not supposed to, and it was not a choice. The dialog renders `{activeTab === 'plan' && <PlanTabContent/>}`, so leaving the tab unmounts it and React drops all fourteen of its `useState` values — preset, direction, conflict policy, verify policy, canary, budget, streams, compression, EC — along with the Sync tab's paths, exclude list and dry-run.

Mounting all three tabs at once would trade this for something worse: the Compare tab starts a recursive scan on mount. So the tabs still unmount and their settings live in a small store owned by the dialog, read back on remount and cleared on each fresh open. Two effects had to learn about it, both of which seed one control from another (speed mode → streams/compression, preset → EC) and therefore also fire on mount, which after a switch is a remount: they would have overwritten exactly the values just restored.

## "dist" was a hash, not a distance (17833081)

Ehud asked to sort the non-identical duplicate groups from most to least similar. Writing that sort is what exposed why it could not have worked: the representative distance was `.max()` of the **signature values** of the group's members, not of the hamming distances between them, on both the raster and the text arm. A pair clustered two bits apart could report a number in the billions. Anything that ordered by it — including the sort he asked for — would have ordered the groups by an arbitrary 64-bit integer. It is now the largest hamming distance from the group's first member, bounded by the threshold that admitted the group. The `other` arm was already computing a real TLSH diff.

The rest of that comment is in too: each group carries its members' own signatures so a row can show why it is in the group (a "Show hashes" toggle keeps them out of the way by default), and the fuzzy cutoff is now the user's to set. It had never left the frontend — `find_duplicate_files` took no distance parameter and always passed `None`, so the per-modality defaults were the only thing anyone could run. There is a Fuzzy cutoff field now, blank meaning those defaults, committed on blur or Enter rather than per keystroke since each commit re-runs the scan.

## The crypt toggle would not go away outside the Overlays Path (17831657 follow-up)

> toggling off the '🔒 RCLONE CRYPT' or '🔒 AEROCRYPT' toggle switch so it looks grey, and then going to the Remote Path directory, the grey toggle switch is still showing.

The badge asks `isWithinCryptScope` which anchor the tab is bound to, and read it from `session.cryptOverlay` — the live vault, which locking clears. The scope then fell back to `''`, and to that predicate `''` does not mean "no anchor", it means "the whole remote is the anchor": every folder in the session read as inside, so the in-scope branch rendered everywhere and the jump-back button EF-02 put in its place could never appear while locked. The anchor now survives the lock the way the overlay kind already did.

## The rest

**Scan progress (17831774, 17832864).** Compare and Find Duplicates each answered a long scan with a bare spinner, the one display that cannot distinguish work from a hang. Compare needed no new backend: every compare walker has emitted `sync_scan_progress` since the recursive scan landed and nothing on the frontend ever listened. Those ticks now also carry directories and bytes. Find Duplicates had nothing to listen to, so the command emits its own — files, directories, bytes and depth during the walk, then a determinate fraction during the analysis pass, throttled to 120 ms so the reporting cannot slow the scan it reports on. Non-identical mode used to let the dedupe engine do its own walk; it now shares exact mode's, so the two cannot drift and both are visible.

**View Trash (17832100, 17832180, 17832268, 17832371).** Four requests on one surface: sorting, a Type column, the Name header lining up with the names instead of reading as if it sat over the checkboxes, timestamps on one line, and rubber-band + Ctrl selection. The obstacle was that '🗑 View Trash' is not one dialog: **seventeen** provider components each carried their own copy of the same table, and `grep sortBy` across all of them returned nothing. Sixty-eight edits and seventeen chances to diverge, or one shared table — this is the shared table, and all seventeen now render it, including the three awkward ones (Internxt read-only, FileLu's `<li>` list, S3's version/delete-marker view with its own Kind and Actions columns).

**Copy buttons in Find Duplicates (17832886).** The `Copy` glyph in that dialog was decorative. Each row now has a real copy button on the file name and on the folder path, plus buttons to open the file in its default app or reveal it in the file manager, so comparing two candidates no longer means reading a path off the screen with an OCR app.

**Encryption badge (17832803).** Add Service has shown "E2E 128-bit"/"E2E 256-bit" on MEGA, Filen and Internxt for a long time; My Servers said it only in prose in the subtitle. Both now read one list and a test holds the Add Service catalogue to it. Not every MEGA protocol qualifies, and the test says so: MEGA S4 is a `protocol: 's3'` profile where the server holds the keys, so badging it would be a false claim about who can read the files.

**TAB.DIGITAL (17831692 follow-up).** The four Ehud still saw were in the VitePress data and config modules, which the 27 July markdown sweep never opened. Fixed in [docs.aeroftp.app#4](https://github.com/axpdev-lab/docs.aeroftp.app/pull/4). The `tab.digital` URLs stay lowercase, including "Sign up at [tab.digital]", where the domain is the link text.

**Drime (17832505 follow-up).** Split into #520, because it is a rename whose 47 locale files pushed this PR one file over CodeRabbit's review limit. Short version: he is right, the catalogue had it inside out, and the generated tables are regenerated rather than hand-edited.

## Verification

`tsc` clean, `npm run build` ok, `vitest` 589/589 (39 new), `cargo clippy --all-features --tests` exit 0 with no warnings, `cargo test --all-features --lib` 3381 passed, `cargo fmt --check` clean, `i18n:validate` 46/46 with 0 missing keys and 0 placeholders.

No new i18n key anywhere in this PR. Every new label reuses one that is already translated in all 47 locales (`contextMenu.copyName`, `contextMenu.copyPath`, `contextMenu.open`, `aeroShare.inbox.revealFile`, `settings.columnType`, `connection.filenEncryptionTitle`), and with the Drime rename split into #520 this PR does not touch the locale files at all.

Two pins were checked against the defect rather than only against the fix: the crypt-scope test fails on the old resolution, and the dedupe distance test fails on the old `.max()` expression.

## Merge order

Rebased onto `main` after #515 landed. It shares `src/App.tsx` with #506 (the crypt badge block), so it should land after #506 and after #516. #520 (the Drime rename, split out of this branch) is independent of both and of this one.

Reported and specified throughout by @EhudKirsh.

None of these close their thread automatically: they stay open for the reporter to confirm.

## Summary by CodeRabbit

* **New Features**
  * Added live duplicate-scan progress (elapsed time + file/dir/byte counters), per-file fuzzy signatures, and richer duplicate results (sorting, optional hash display, and quick actions).
  * Added fuzzy cutoff controls and improved directory-compare progress metrics.
  * Standardized trash tables across providers with shared sorting/selection; added native E2E encryption badges for saved services.
  * Preserved Plan/Sync tab state and improved crypt-scope handling for locked vs unlocked behavior.
* **Bug Fixes**
  * Fixed fuzzy similarity “distance” to use true Hamming-distance and ensured the cutoff/threshold behavior is honored.


axpnet:
All ten addressed in `12be1292b`, none dismissed. Two of them were defects in this PR's own new code, one is a performance mistake in it, one predates it, and the rest are correctness details in the new UI.

**`force` never passed `true`** — right, and the consequence is worse than a missing frame: the throttle could swallow the final tick, so the bar stayed short of the total for as long as the dialog stayed open. Both modes now force a tick when the pass ends.

**Fuzzy path has no I/O budget** — right, and it predates this PR: the engine buffered every candidate's full contents in a `Vec` before computing any signature. It now reads and signs one file at a time, dropping the bytes as soon as the signature is out, under the same 2 GB budget exact mode uses. Reading one large file at a time is bounded; holding a directory of them was not.

**Always-zero clone-pool `dirs_found`** — right, and the fix is to say nothing rather than to find a number: that branch returns a flat file list, so there is no directory count to report on that path. The field is omitted and the frontend reducer keeps the last value it saw, which is the behaviour its test already pins.

**O(D×N) rescan** — right. Running counters now, adjusted at the insert site by whatever the insert replaced, since the walk can revisit a path (which is why it was recomputing in the first place).

**Ref mutated during render** — right, and the constraint that led me there was real: the store has to be empty *before* a tab mounts and reads it, and an effect runs after. Resetting on **close** satisfies it without writing a ref mid-render, since that is the same moment from the store's point of view.

**Hardcoded English labels** — right. Five keys added and translated across all 47 locales (`duplicates.sortWasted`, `sortSimilarity`, `showHashes`, `fuzzyCutoff`, `fuzzyCutoffHint`), as one-line insertions so each catalogue keeps its formatting. The `<label>` with nothing to point at is fixed by passing the text through the Checkbox's own `label` prop, which also gives it the hit area the cursor was promising.

**FileLu null `file_code`** — right, and it is three bugs in one: duplicate React keys, one selectable row standing for many, and a per-row action invoked with an empty code. Filtered out, matching the two `.filter(Boolean)` call sites already in that file.

**FileLu per-row actions** — right, and inherited: I lifted those inline handlers out of the old list unchanged. Both now share the bulk handlers' shape, with an activity-log entry that settles and a failure the user can see.

**Nextcloud `deleted_at === 0`** — right: displayed as an em dash, sorted as 1970. Mapped to null so the sort agrees with the display.

**`size: null` renders `0 B`** — right, and `sortTrashRows` already disagreed with the cell. Null is now a dash for rows that have no size, not only for folders.

Re-verified after the changes: `tsc` clean, `vitest` 589/589, `i18n:validate` 47/47 with 0 missing keys and 0 placeholders, `cargo clippy --all-features --tests` with no warnings, `cargo fmt --check` clean, dedupe tests green.

axpnet:
The `File chooser is portal-backed` red on this branch was timing, not code: a re-run of the **same commit** (`d02bcdaf6`) passed. The app hit its own 10-second splash safety timeout before the frontend signalled `app_ready`, so the harness never drove the UI and eleven downstream assertions failed after it; the frontend server had answered 200, both `NetworkMonitor` and `ProxyResolver` were asked and answered, and scenarios 2 and 3 started the app fine in the same job. `cancel.log` reaches the same conclusion on its own: *"the app never reported app_ready, so the UI was never driven. This is an environment failure, not a chooser failure."*

Filed as #521 rather than patched here: the gate has no readiness budget of its own and inherits a timeout that exists for a different purpose, but making a gate more forgiving is not a change to slip into an unrelated PR.

axpnet:
Correction to my note above, since it named the wrong cause and therefore the wrong fix. I wrote that the gate "has no readiness budget of its own" and inherits the app's 10-second splash timeout. It does have one: `recon-session.sh` waits up to `RECON_WAIT_SECONDS`, default **45 seconds**, for `app_ready`, and already fails closed with a single named precondition error.

The timeline says the app started at 03:52:48 and the harness reported `app_ready seen: 0` at 03:53:39, so it used that full budget. The app did not miss a 10-second deadline: it never rendered at all in 45 seconds, with its own frontend server answering 200, which per the harness's own comment means WebKit did not render a bundle that was served fine. Still environmental, still did not reproduce on the re-run, but the fix is not a longer wait.

What is left is the cascade: after `run_case` returns non-zero, `portal-chooser-test.sh` keeps evaluating ten more assertions against evidence the failed session could not produce, so a rendering hiccup reads as "the chooser did NOT go out of process". #521 is updated with the corrected diagnosis and a way to reproduce the shape deliberately (`RECON_WAIT_SECONDS=1`) instead of waiting for a bad runner.

PR #523, opened 2026-07-30
test(portal): stop chooser failure cascades

### LINKED ISSUE #521: Portal file chooser gate: a cold start over 10s fails as eleven chooser assertions
The portal gate failed once on #519 and passed on a re-run of the **same commit** (`d02bcdaf6`), so this is not a code regression. Recording it because the failure is loud, eleven assertions red, and every one of them accuses the file chooser, while the chooser was never reached.

> **Correction, 2026-07-30.** The first version of this issue said the gate "has no readiness budget of its own" and inherits the app's 10-second splash timeout. That is wrong, and it matters, because it pointed at a fix that is already implemented. `recon-session.sh` waits up to `RECON_WAIT_SECONDS`, **default 45 seconds**, for `app_ready` in the app log, and already fails closed with a single named precondition error. The corrected reading is below.

## What actually happened

Run [30511254611](https://github.com/axpdev-lab/aeroftp/actions/runs/30511254611), scenario 1 (`cancel`):

```
03:52:48  app started (pid 36469)
03:52:59  Splash screen safety timeout reached, force-closing
03:53:39  app_ready seen: 0        <- after the harness's full 45s budget
```

So the app did not miss a 10-second deadline. It **never rendered at all**, in 45 seconds, in that session. The splash timeout at 10 seconds is a symptom logged on the way past, not the deadline that decided the outcome.

The harness's own discriminator, which exists for exactly this, answered clearly: the app's frontend server returned `http_status=200 bytes=1263`, so the bundle was served and the server was up. Per `recon-session.sh`'s comment, that combination means "WebKit did not render it (sandbox, dmabuf, GL under Xvfb)", not "the frontend was never embedded". `NetworkMonitor` and `ProxyResolver` were both asked and both answered, ruling out the documented "an incomplete stand-in becomes a portal that breaks the app" case. Scenarios 2 and 3 started the app fine in the same job.

The run was loaded: four workflows on one push, plus the build jobs of a merge to `main` in parallel.

## The defect this leaves behind

Not the readiness wait, which is correct. The problem is what `portal-chooser-test.sh` does **after** a session fails its precondition:

```
rc=$(run_case cancel RECON_MODE=cancel)
[ "$rc" = "0" ] && ok "the session ran to completion" || bad "the session failed (exit $rc), ..."
grep -q "windows with class aeroftp on" ... || bad "the app did not open on the private display"
[ "$(calls cancel)" -ge 1 ] || bad "the portal was never asked: the chooser did NOT go out of process"
call1 cancel | grep -q '"method":"OpenFile"' || bad "no OpenFile call recorded"
...
```

`run_case` returned 1 and said why, and the script then evaluated ten more assertions against evidence that the failed session could not have produced. Every one of them is *technically true*, and every one of them is *misleading*: "the chooser did NOT go out of process" reads as a chooser regression when the chooser was never invoked. A reader who trusts the summary line, `passed: 10 failed: 11`, learns the wrong thing about their own diff.

## What to change

The eleven-red cascade is the bug. Two ways, and they compose:

1. **Stop the cascade.** When `run_case` returns non-zero, report that one failure and skip the assertions that depend on the session having run, rather than emitting them as chooser failures. A skipped assertion is honest; a failed one is a claim.
2. **Retry a precondition failure once.** The rendering failure above is environmental and did not reproduce. One retry of the scenario, with the retry recorded in the output so it never becomes invisible, keeps the gate meaningful without failing the build on a WebKit hiccup under Xvfb.

Not proposed: raising the app's splash timeout, which exists for a user-facing reason and had no bearing here.

## How to verify a fix

The whole gate needs a Tauri build, so iterate on the harness alone:

```bash
# the stand-in's own tests, seconds, no app build
tests/portal-chooser/selftest-portal.sh

# the full gate, needs the release binary built by the Tauri CLI (never a bare cargo build)
tests/portal-chooser/portal-chooser-test.sh src-tauri/target/release/aeroftp
```

To exercise the failure path deliberately without waiting for a slow runner, force the precondition to fail and check the output shape:

```bash
RECON_WAIT_SECONDS=1 tests/portal-chooser/portal-chooser-test.sh src-tauri/target/release/aeroftp
```

A fixed gate reports **one** failure naming the precondition, plus skips, instead of eleven chooser assertions.

## Done when

- [ ] A session that fails its precondition produces one failure and no chooser assertions.
- [ ] `RECON_WAIT_SECONDS=1` demonstrates that shape, so the behaviour is pinned by something runnable rather than by a runner's bad day.
- [ ] A retry, if added, is visible in the log when it happens.

Found while landing #519 (the discussion #347 wishlist batch). Harness delivered in #505 for #464.

axpnet:
## Summary

- retry an `app_ready` environmental precondition failure once, with a visible `RETRY` line
- preserve the first attempt's evidence beside the final attempt
- stop each failed session at one primary failure and mark dependent chooser assertions as skipped
- include skipped assertions in the final gate summary

Closes #521.

## Verification

- `npm run tauri build`
- `tests/portal-chooser/selftest-portal.sh` (`13 passed, 0 failed`)
- `RECON_ALLOW_WAYLAND_HOST=1 tests/portal-chooser/portal-chooser-test.sh src-tauri/target/release/aeroftp` (`21 passed, 0 failed, 0 skipped`)
- `RECON_ALLOW_WAYLAND_HOST=1 RECON_WAIT_SECONDS=1 tests/portal-chooser/portal-chooser-test.sh src-tauri/target/release/aeroftp` (expected non-zero: `3 failed, 18 skipped`, one named session failure per scenario, visible retry, no dependent chooser failures)
- `bash -n tests/portal-chooser/portal-chooser-test.sh`
- `git diff --check`

### DIFF
--- a/tests/portal-chooser/portal-chooser-test.sh
+++ b/tests/portal-chooser/portal-chooser-test.sh
@@ -35,17 +35,37 @@ SEQ="${PORTAL_TEST_SEQ:-Export / Import|Import Servers Restore from .aeroftp bac
 
 PASS=0
 FAIL=0
+SKIP=0
 ok()  { printf '  ok   %s\n' "$1"; PASS=$((PASS + 1)); }
 bad() { printf '  FAIL %s\n' "$1" >&2; FAIL=$((FAIL + 1)); }
+skip() { printf '  skip %s\n' "$1"; SKIP=$((SKIP + 1)); }
 
 run_case() {
   local name="$1"; shift
   local out="$WORK/$name"
+  local rc
   rm -rf "$out"
   # shellcheck disable=SC2086
   env "$@" RECON_OUT="$out" RECON_PRESS="$SEQ" RECON_POST_SETTLE=6 \
       "$HERE/recon-session.sh" "$APP" >"$WORK/$name.log" 2>&1
-  echo "$?"
+  rc=$?
+
+  # A renderer that misses app_ready is an environmental precondition failure,
+  # not evidence about the chooser. Retry that one failure once, visibly, while
+  # preserving the first attempt for diagnosis. Other harness failures are not
+  # retried: repeating a malformed setup would only hide the real error.
+  if [ "$rc" -ne 0 ] &&
+      grep -q "::error::the app never reported app_ready" "$WORK/$name.log"; then
+    mv "$out" "$WORK/$name-attempt1"
+    mv "$WORK/$name.log" "$WORK/$name-attempt1.log"
+    printf '  RETRY %s missed the app_ready precondition; retrying once\n' "$name" >&2
+    # shellcheck disable=SC2086
+    env "$@" RECON_OUT="$out" RECON_PRESS="$SEQ" RECON_POST_SETTLE=6 \
+        "$HERE/recon-session.sh" "$APP" >"$WORK/$name.log" 2>&1
+    rc=$?
+  fi
+
+  echo "$rc"
 }
 
 # `grep -c` prints 0 AND exits 1 when a file has no matching lines, so the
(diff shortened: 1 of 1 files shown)


axpnet:
@coderabbitai review

PR #529, opened 2026-07-30
fix(sync): name the export action, and stop shipping a raw i18n key

axpnet:
Two defects in the AeroSync Templates dialog, both on screen, both found while verifying #514 in the running app rather than by reading the code.

## The raw key

The Bash format radio shipped as `Bash script (.sh)(common.default)`.

`common.default` was never a key in `en.json`, and the call site's guard could not save it:

```tsx
({t('common.default') || 'default'})
```

`t()` returns the key itself when it misses, and a non-empty string is truthy, so the `||` branch was unreachable by construction. The app logged `[i18n] Missing translation: common.default` on every render of the dialog while the user read the key on screen.

The key is added and translated across all 46 non-English locales, and the dead guard is removed rather than left in place implying a safety it never provided.

## The button that named a format instead of an action

`exportButtonLabel` fell back to `syncPanel.templateFormatBash` and `syncPanel.templateFormatPwsh` for the two script formats. Those are the exact strings used by the format radios, so the dialog presented two identical `Bash script (.sh)` controls, and the only one that did anything read as one more format to pick.

It now reads `Export script` for the three script formats and `Export Template` for `.aerosync`. Both are keys that already existed and are already translated, so this adds no new string beyond `common.default`.

## Verified in the app

Driven headlessly through the WebKit inspector against a connected SFTP remote, reading the labels back from the live DOM:

| Format selected | Action button |
|---|---|
| `.aeroftp-script` | `Export script` |
| `.aerosync` | `Export Template` |
| Bash | `Export script` |
| PowerShell | `Export script` |

and the Bash radio now reads `Bash script (.sh)(Default)`.

## Gates

`tsc --noEmit` clean, `vitest` 592/592, `vite build` clean, `i18n:validate` 47/47 with 0 placeholders and 0 orphan keys. No Rust touched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Export actions now display the correct label based on the selected format.
  * Format options consistently show the localized “Default” hint.

* **Localization**
  * Added localized “Default” labels across supported languages.

### DIFF
--- a/src/components/Sync/SyncTemplateDialog.tsx
+++ b/src/components/Sync/SyncTemplateDialog.tsx
@@ -392,14 +392,14 @@ export const SyncTemplateDialog: React.FC<SyncTemplateDialogProps> = ({
 
     if (!isOpen) return null;
 
+    // The button names the action, never the format. Labelling it
+    // `templateFormatBash` put a second control reading `Bash script (.sh)`
+    // next to the radio of the same name, so the only button that does
+    // anything read as one more format to choose from.
     const exportButtonLabel =
         exportFormat === 'aerosync'
             ? t('syncPanel.templateExport')
-            : exportFormat === 'aeroftp-script'
-                ? t('syncPanel.aerosyncScriptExportButton') || 'Export script'
-                : exportFormat === 'bash'
-                    ? t('syncPanel.templateFormatBash')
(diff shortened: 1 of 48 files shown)


axpnet:
@coderabbitai full review

axpnet:
@coderabbitai review

PR #530, opened 2026-07-30
fix(sync): stop the script exports dropping the preset's excludes

axpnet:
Found while verifying #514 in the running app, on the same preset, exported twice.

`.aeroftp-script` came out with `node_modules`, `.git`, `.DS_Store`, `Thumbs.db`, `__pycache__` and `target`. The same Mirror preset exported as bash came out with none.

## Why

The dialog's exclude list is empty whenever the user has not typed a pattern of their own. `export_sync_script_cmd` and `export_sync_template_cmd` took a plain `Vec<String>` and used it as given, so an empty list read as "exclude nothing" rather than "the user added nothing". `aerosync_export_script_cmd` never had the bug: it takes an override of `Option<Vec<String>>` and falls back to the preset when it is `None`.

This is not a cosmetic difference between formats. A Mirror script that has quietly lost its excludes does not fail, it walks straight into `.git` and `node_modules` and uploads them.

## The fix

One rule, in one place:

```rust
pub fn resolve_exclude_patterns(
    caller_patterns: Option<Vec<String>>,
    preset_patterns: &[String],
) -> Vec<String>
```

A non-empty caller list wins, anything else means the preset's own. All three export commands go through it, and the two that took `Vec<String>` now take `Option<Vec<String>>` so the frontend can express "nothing of mine" instead of "exclude nothing". The two call sites use the expression the `.aeroftp-script` path already used.

## Verified against the running backend

Both commands invoked directly through the WebKit inspector, on the Mirror preset:

| Call | bash script | `.aerosync` template |
|---|---|---|
| `exclude_patterns: null` | `--exclude 'node_modules'`, `'.git'`, `'.DS_Store'`, `'Thumbs.db'`, `'__pycache__'`, `'target'` | same six |
| `exclude_patterns: ['*.tmp']` | `*.tmp` only | `*.tmp` only |

Three unit tests pin the helper, including the case that caused this: `Some(vec![])` resolves to the preset, not to nothing.

## Gates

`cargo fmt`, `cargo clippy --all-features --tests` clean, `cargo test --lib` 3385 passed with 0 failed, `tsc --noEmit` clean, `vitest` 592/592.

Found while working #514.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sync template and script exports now automatically use the preset’s exclude patterns when no custom patterns are provided.
  * Empty exclude-pattern selections are handled consistently, while explicitly chosen patterns continue to override preset defaults.
  * Improved export behavior for presets without exclude patterns.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -13090,7 +13090,9 @@ fn export_sync_template_cmd(
     profile_id: String,
     local_path: String,
     remote_path: String,
-    exclude_patterns: Vec<String>,
+    // `None` (or an empty list) means "the preset's own", see
+    // `sync::resolve_exclude_patterns`.
+    exclude_patterns: Option<Vec<String>>,
 ) -> Result<String, String> {
     let profiles = sync::load_sync_profiles()?;
     let profile = profiles
@@ -13103,13 +13105,14 @@ fn export_sync_template_cmd(
     } else {
         None
     };
+    let excludes = sync::resolve_exclude_patterns(exclude_patterns, &profile.exclude_patterns);
     let template = sync::export_sync_template(
         &name,
         &description,
         profile,
         &local_path,
         &remote_path,
-        &exclude_patterns,
+        &excludes,
         schedule_opt,
     )?;
     serde_json::to_string_pretty(&template).map_err(|e| e.to_string())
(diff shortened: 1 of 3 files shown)


axpnet:
@coderabbitai full review

PR #531, opened 2026-07-30
fix(cli): stop find dropping --name and the default glob under --profile

axpnet:
## What was wrong

`aeroftp-cli find` dropped `--name` and the default `*` glob whenever `--profile` was used, reporting `status: ok` with zero matches on a directory full of files.

With `--profile` the positional arguments shift left: the first positional holds the path and the second holds the pattern. The dispatch branch for that case passed `path` through as the pattern without ever reading the resolved pattern, so:

- `--name` was ignored entirely.
- With the pattern omitted, the search ran against `path`'s own clap default, `/`.

A glob is matched against the entry name (`matcher.is_match(&e.name)` in `cmd_find`), and no file name contains `/`, so the search matched nothing. The command still exited `0` with `"status": "ok"`, which makes it a silent wrong answer rather than a visible failure: an agent reading the JSON concludes the directory is empty.

## Fix

The dispatch now routes both branches through a pure `resolve_find_target(has_profile, url, path, pattern, name)` helper. In the shifted case `--name` still wins, and an omitted pattern falls back to the documented `*` instead of `/`. The clap `default_value`s for `path` and `pattern` were changed to reference the same `FIND_DEFAULT_PATH` / `FIND_DEFAULT_PATTERN` consts the fallback compares against, so the parser default and the fallback cannot drift apart.

The other four sites with a multi-positional shift (`edit`, `versions get/restore/purge`) were checked: they perform the shift correctly, `find` was the only broken one.

## Test evidence

Unit tests added in `src-tauri/src/bin/aeroftp_cli.rs`:

- `find_with_a_profile_keeps_the_pattern_instead_of_the_path_default`: covers the shifted case (pattern omitted falls back to `*`, `--name` honoured, shifted positional used as-is, `--name` beating the shifted positional) and the non-shifted cases (no profile, profile with an explicit URL, profile with the `_` placeholder).
- `find_positional_defaults_match_the_shift_fallback`: parses `aeroftp-cli --profile 42 find /base` with clap and asserts the omitted pattern really arrives as `path == "/"`, guarding the assumption the fallback rests on.

Red-if-reverted check: restoring the old `("_", url, path)` branch makes the first test fail with

```
assertion `left == right` failed
  left: ("_", "/Nas_Prog/inbox/", "/")
 right: ("_", "/Nas_Prog/inbox/", "*")
```

Gates run on the branch:

```
cargo fmt --all                                        clean
cargo clippy --bin aeroftp-cli --tests -- -D warnings  clean
cargo test --bin aeroftp-cli                           497 passed, 0 failed
```

## Live validation

Profile "SSH MyCloud HD" (SFTP), a directory holding 406 `.md` files out of 500 entries. Same profile, same path, before and after.

| Invocation | Before | After |
|---|---|---|
| `find <path> "*.md"` (positional) | 406 | 406 |
| `find <path> --name "*.md"` | 0 | 406 |
| `find <path>` (default glob) | 0 | 500 |

Note for whoever repeats this: address the profile by NAME, not by index. The installed binary and the freshly built one resolved index 42 to two different profiles, which looks exactly like a regression and is not one.

No linked user issue: the bug surfaced in internal dogfood.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `find` command argument handling when using profiles.
  * Omitted search patterns now correctly default to `*` instead of `/`.
  * Explicit `--name` values are now honored consistently.
  * Default path and pattern behavior is aligned across command-line parsing and execution.

### DIFF
--- a/src-tauri/src/bin/aeroftp_cli.rs
+++ b/src-tauri/src/bin/aeroftp_cli.rs
@@ -1680,11 +1680,11 @@ enum Commands {
         #[arg(default_value = "_", hide_default_value = true)]
         url: String,
         /// Base path to search from
-        #[arg(default_value = "/")]
+        #[arg(default_value = FIND_DEFAULT_PATH)]
         path: String,
         /// Search pattern (glob-style). Positional; `--name` is an
         /// equivalent flag form for agents that expect named args.
-        #[arg(default_value = "*")]
+        #[arg(default_value = FIND_DEFAULT_PATTERN)]
         pattern: String,
         /// Glob pattern alias for the positional argument. When both
         /// the positional pattern and `--name` are set, `--name` wins.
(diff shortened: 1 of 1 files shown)


axpnet:
@coderabbitai review

PR #532, opened 2026-07-30
fix(commands): drain the rest of the main-thread commands out of lib.rs (82 to 23)

### LINKED ISSUE #517: Synchronous Tauri commands run on the main thread: audit the 37 that are still sync
## What this is

A follow-up opened from #510 / #515, where the same defect was found and fixed in one command. It is filed separately because the audit is a different piece of work from the picker fix, not because it belongs to someone else.

Tauri v2 runs a **synchronous** `#[tauri::command]` on the **main thread**; only an `async fn`, or `#[tauri::command(async)]`, moves it elsewhere. On Linux the main thread is the GTK thread, so any blocking work inside a sync command freezes the whole window for its duration, with no spinner and no way for the user to tell it apart from a hang.

This repo pins `tauri = "=2.11.0"` and has, at the time of writing, **37 synchronous commands against 573 async ones**. The convention is therefore already "async", and `tokio::task::spawn_blocking` is already used 112 times for exactly this shape of work. The question is which of the 37 are sync by accident.

## Measured, not assumed

Fixed in #515, listed here as the pattern to look for:

- `portal_chooser::chooser_unavailable` was sync while doing `gio::bus_get_sync` followed by up to two `call_sync` round trips at 1000 ms each. Worst case was roughly two seconds of frozen window in front of a user gesture, plus an unbounded tail if the bus socket accepts a connection and never completes authentication. It is now `async` with the blocking round trip on `spawn_blocking`, pinned by a test whose `.await` fails to **compile** if anyone makes it sync again.

Two more found while grepping for the same shape, **not** fixed, and the reason this issue exists:

- **`filesystem.rs`, `list_subdirectories(path: String)`** does `std::fs::read_dir(dir_path)` on the main thread, on a path that comes from the frontend. On a healthy local disk this is microseconds. On an unresponsive network mount (NFS, SSHFS, a disconnected SMB share) `read_dir` blocks in the kernel for as long as the mount takes to time out, which can be minutes and is not bounded by anything in our code. This is the worst of the three because the input is user-supplied and a file manager is exactly the app where someone browses into a stale mount.
- **`settings.rs`, `native_rsync_classic_available`** walks `PATH` looking for the `rsync` binary. Bounded and fast in normal conditions, but every entry is a filesystem metadata call, so a `PATH` containing a dead mount stalls the same way.

## What closing this looks like

1. Enumerate the 37 sync commands, e.g. by diffing `#[tauri::command]` occurrences against the `pub async fn` ones, and record the list in the issue so the audit is checkable rather than claimed.
2. For each, classify: pure computation and small state reads may stay sync; anything doing filesystem, network, D-Bus, subprocess or lock-waiting work becomes `async` plus `spawn_blocking`.
3. Fix the ones that qualify. `list_subdirectories` first, since its input is user-controlled and its stall is unbounded.
4. Pin the class rather than the instances, so this cannot come back by addition. The `.await` trick used in #515 pins one command; a stronger version is a test that walks the source and fails when a command in a declared blocking-work allowlist is declared sync, which is the same shape as `pickPathIsTheOnlyPicker.test.ts` in #515 and works for the same reason: the next new command becomes red instead of silently reintroducing the freeze.

## Why it matters beyond the freeze

A frozen window is reported as "the app hangs when I click X", which is a bug report nobody can localise, and on Linux under WebKitGTK a long main-thread block is also what a compositor reports as unresponsive. The picker case in #510 was found only because a gate was measuring something adjacent. Nothing is currently watching for the rest.

axpnet:
Closes #517. Follows #524, which did the first half and left the second in a list that could only shrink.

## What this empties

#524 moved the commands that do filesystem, keystore and clipboard work outside `lib.rs`, and froze the remaining 74 in `MAIN_THREAD_NOT_YET_MOVED` — a ratchet, asserted by the class pin, that a new synchronous command could not join and that converting one without deleting its line would break. This drains it, so the constant is deleted rather than carried.

**59 more commands become `async` with their body on the blocking pool. 15 stay synchronous and are now in `MAIN_THREAD_ALLOWED` with the reason next to each.** From 82 out of 853 at the start of this work to 23.

## The 59

The bulk is one family: sync index, journals, profiles, snapshots, templates, scripts, cloud config, cloud pairs, versioning. All of it JSON read and write against the config directory, all of it previously on the GTK thread. Three are worth naming because they are not short at all:

- **`versions_disk_usage`** walks a directory tree, and its cost grows with the user's own version history;
- **`verify_local_transfer`** hashes a file;
- **`safe_picker_start_dir`** walks *up* a path supplied by the frontend with an `is_dir()` per level, so an unreachable mount stalls it with no bound of ours — the same shape as `list_subdirectories` in #524.

Plus `copy_to_clipboard`, `get_system_info` (stats the vault and known_hosts), `portable_info`, `flatpak_config_import_*` and `mount_autostart_blocked`.

## The 15, and why each is not a dodge

Three **must** be on the main thread, and saying so is the correct answer rather than a concession: `restart_app` drives the event loop, `toggle_menu_bar` calls GTK window operations, and `rebuild_menu` already marshals menu construction onto the main thread through a channel — which `tauri-runtime-wry`'s `send_user_message` runs *inline* when the caller is already that thread, so making it async would add an event-loop round trip and move no work at all. Same reasoning that keeps `aeroshare_notify` where it is.

The other twelve are bounded by this crate rather than by the caller: atomic loads and stores, `default()` structs, version strings baked in by `build.rs`, argv and environment reads, one HMAC, and string classification of an error we produced ourselves.

**One candidate did not survive its own test.** `get_default_cloud_folder` was on the list to justify: it looks like pure construction. But `CloudConfig::default()` calls `dirs::document_dir()`, which on Linux reads `~/.config/user-dirs.dirs`. That is disk, so by the bar written above `MAIN_THREAD_ALLOWED` — *bounded by our own code, waits on nothing outside the process* — it gets converted like the rest. A bar that bends for the case in front of you is not a bar.

## The pin, after

`MAIN_THREAD_NOT_YET_MOVED` is gone because it is empty, which is the only good reason to delete a ratchet. What remains is the single allowlist and set equality in both directions: a new synchronous command is red, and an entry whose command became async, was renamed or was deleted is also red. `every_allowlist_entry_explains_itself` keeps a bare name from being a way to get green — and it caught three of my own entries during this work, which were too short to be reasons; they were expanded rather than the assertion loosened.

## Two mechanical notes, both from getting them wrong first

The conversion was scripted, and the script was wrong twice in ways worth writing down:

1. it took the return type with `rindex(')')`, which picks the paren inside `Result<(), String>` and silently mangles every command that returns unit. It matches the parameter list by depth now. Caught because the first run aborted on `copy_to_clipboard`, not because the output was reviewed;
2. a zero-argument wrapper must pass the function rather than `move || f()`, which is `clippy::redundant_closure`, and a plain-value fallback must bind `err` or `-D warnings` rejects it. Both are generated correctly rather than patched afterwards, because the first attempt at patching afterwards used a multi-line regex that matched far too greedily and corrupted the file; that was reverted from git rather than repaired by hand.

`verify_local_transfer` fails closed on the JoinError path, with `passed: false` and a message, because a verification that never ran did not pass.

## Internal callers, fixed rather than worked around

`lib.rs`'s own `safe_picker_start_dir_tests` drove the command synchronously and now drive the blocking body directly. `PortableInfo` and `SystemInfo`, both private to `lib.rs`, derive `Default` so the fallback is a value rather than a hand-written literal that would drift from the struct the first time somebody adds a field.

## Verification

Gate below in a comment, with the exit codes read without a pipe.

## Summary by CodeRabbit

* **Performance & Reliability**
  * Improved responsiveness by running sync, cloud, file, clipboard, and system operations off the main thread.
  * Standardized background failure handling with safer fallbacks.
  * Transfer verification now fails closed if verification can’t complete.
* **Bug Fixes**
  * Reduced chances of lost or corrupted updates when multiple cloud/multi-path configuration changes occur.
* **Tests**
  * Updated main-thread execution checks and allowlists to ensure blocking work isn’t run synchronously.

### DIFF
--- a/src-tauri/src/cloud_config.rs
+++ b/src-tauri/src/cloud_config.rs
@@ -238,6 +238,10 @@ fn get_config_path() -> PathBuf {
 
 /// Load cloud configuration from disk
 pub fn load_cloud_config() -> CloudConfig {
+    load_cloud_config_unlocked()
+}
+
+fn load_cloud_config_unlocked() -> CloudConfig {
     let config_path = get_config_path();
 
     if config_path.exists() {
@@ -257,12 +261,35 @@ pub fn load_cloud_config() -> CloudConfig {
     CloudConfig::default()
 }
 
+/// Load → mutate → save under one write lock.
+///
+/// Use this for every read-modify-write of `cloud_config.json`. Taking the lock
+/// only around `save_cloud_config` still allows two concurrent mutators on the
+/// blocking pool to both load the same snapshot and last-write-win (lost
+/// update). The main-thread command path used to serialise this for free;
+/// `spawn_blocking` does not.
+pub fn with_cloud_config_mut<F, T>(f: F) -> Result<T, String>
+where
+    F: FnOnce(&mut CloudConfig) -> Result<T, String>,
+{
+    let _lock = CONFIG_WRITE_LOCK
+        .lock()
+        .map_err(|_| "Config write lock poisoned".to_string())?;
+    let mut config = load_cloud_config_unlocked();
+    let out = f(&mut config)?;
+    save_cloud_config_unlocked(&config)?;
+    Ok(out)
+}
+
 /// Save cloud configuration to disk (serialized via write lock, atomic temp+rename)
 pub fn save_cloud_config(config: &CloudConfig) -> Result<(), String> {
     let _lock = CONFIG_WRITE_LOCK
         .lock()
         .map_err(|_| "Config write lock poisoned".to_string())?;
+    save_cloud_config_unlocked(config)
+}
 
+fn save_cloud_config_unlocked(config: &CloudConfig) -> Result<(), String> {
     let config_path = get_config_path();
 
     // Ensure parent directory exists
--- a/src-tauri/src/cloud_pairs.rs
+++ b/src-tauri/src/cloud_pairs.rs
@@ -165,6 +165,10 @@ fn get_pairs_path() -> PathBuf {
 
 /// Load pairs config (absent file => empty pairs, back-compat path).
 pub fn load_cloud_pairs_config() -> CloudPairsConfig {
+    load_cloud_pairs_config_unlocked()
+}
+
+fn load_cloud_pairs_config_unlocked() -> CloudPairsConfig {
     let path = get_pairs_path();
     if path.exists() {
         match fs::read_to_string(&path) {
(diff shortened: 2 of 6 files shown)


axpnet:
@coderabbitai review

axpnet:
@coderabbitai review

axpnet:
### CodeRabbit follow-up (003ae6b6b)

All 7 load-modify-save races (1 inline + 6 outside-diff) were the same shape and are fixed together:

- `cloud_config::with_cloud_config_mut` / `cloud_pairs::with_cloud_pairs_mut` / `sync::with_multi_path_config_mut` hold the write lock across load+mutate+save
- GUI mutators converted: path pairs add/remove, cloud pairs add/remove/update, excluded folders, conflict strategy, enable/pause/resume
- Background worker `last_sync` stamp now RMW under the same lock so it cannot clobber a concurrent GUI field update

@coderabbitai review

PR #539, opened 2026-07-31
fix(providers): six defects found by a cross-provider listing sweep, four of them able to make sync --delete remove the user's local files

axpnet:
Two provider XML defects found while running the post-merge live proof of the LIVE-1 trim_text fix (PR #536) against merged `main` (`3da8058f9`). Neither is a regression of that work: both predate it.

## 1. Azure Blob: any nested directory listed as EMPTY, and `sync --delete` then deleted local files

`parse_blob_list` stripped the provider's `current_prefix` from every `<Name>` Azure echoes. Azure echoes names as full paths from the container root, and the prefix that makes them relative is the one the **request** carried, not the provider's cwd. On any one-shot listing (CLI, MCP, sync's remote scan, the first GUI call after connect) `current_prefix` is still empty, so each blob name kept its full path, the `!relative.contains('/')` depth filter dropped every file, and the directory came back empty. Listed subdirectories kept a full-path name too, so descending into one landed on a path that does not exist.

This is data loss, not a display glitch. `sync --direction download --delete` against such a directory sees an empty remote and plans `delete_local` for every local file, exactly the cascade the LIVE-1 audit was about.

Live reproduction and fix, profile `AzureBlob` (`aeroftp2026/aerocontainer`), four files uploaded to `/aerocontainer/kimi-mainproof`:

```
BEFORE (CLI built from merged main 3da8058f9)
  stat  /aerocontainer/kimi-mainproof/plain.bin  -> found, 1048576 bytes
  ls    /aerocontainer/kimi-mainproof            -> (empty directory), 0 items
  sync --direction download --delete --dry-run   ->
    Sync plan: 0 upload, 0 download, 4 delete, 0 rename, 0 conflict-rename, 0 skipped
      DELETE (local)  plain.bin
      DELETE (local)  sp ace & ünïcodé.txt
      DELETE (local)  raw & name2.txt
      DELETE (local)  a& &b.txt

AFTER (this branch)
  lsjson /aerocontainer                          -> 'kimi-mainproof'  (was 'aerocontainer/kimi-mainproof')
  lsjson /aerocontainer/kimi-mainproof           -> all four names, byte-exact
  sync --direction download --delete --dry-run   ->
    Sync plan: 0 upload, 4 download, 0 delete, 0 rename, 0 conflict-rename, 0 skipped
```

The fixture is the LIVE-1 one, so it doubles as a check that entity handling still holds on Azure: `a& &b.txt` lists exactly.

Introduced no later than v3.5.6 (`380eee934`, 2026-04-19).

## 2. STS: error messages truncated at the last XML entity

`parse_sts_error` assigned each Text fragment to `code`/`message` instead of appending, and handled no `GeneralRef` at all. quick-xml reports `role &quot;dev &amp; ops&quot;` as Text / GeneralRef / Text, so every entity both vanished and reset the field, leaving only the tail after the last one. On the pre-fix parser the new test returns `Some("shared")` for a message whose real text is `Not authorized for role "dev & ops" on bucket <shared>`: the operator was shown one word instead of the reason the assume-role call failed.

Same shape as the LIVE-1 fix: `trim_text(false)`, accumulate Text and GeneralRef, trim once at consumption. `parse_assume_role_response` is deliberately left alone, its fields are base64 credentials and an RFC3339 timestamp and cannot contain entities.

## Tests

New:
- `providers::azure::tests::parse_blob_list_nested_listing_ignores_current_prefix` (parses a nested page with the requested prefix, then pins the old empty-directory behaviour as the broken case)
- `providers::sts::tests::parses_sts_error_message_with_entities`

Renamed to say what it locks: `parse_blob_list_strips_current_prefix_and_reports_last_page` -> `parse_blob_list_strips_listed_prefix_and_reports_last_page`.

Both new tests were run against the pre-fix parsers and observed failing before being kept.

Gates: `cargo test --lib providers::` 819 passed, 0 failed, 1 ignored. `cargo clippy --all-targets --features aerorsync -- -D warnings` clean, also under `--cfg ci_lane3`. `cargo fmt --all --check` clean. Release CLI rebuilt from this branch for the live run above.

---

## Update: a cross-provider sweep, and four more defects it proved

After the two fixes above, the same fixture was run against **every writable saved profile** (72 of 96; live websites, repo-backed GitHub profiles, public read-only test servers and the owner's NAS were excluded on purpose). For each profile: upload the four files, read the names back **from the listing**, fetch each file **by the listed name** and compare sha256, then `sync --direction download --delete --dry-run` and require zero deletes.

Result: **36 providers pass end-to-end**, 17 were unreachable for expired credentials or a down service, 2 refused the fixture by provider policy (Cloudinary rejects `.bin`, Immich accepts media only), and the rest exposed the following.

### 3. Every strict S3 endpoint could not create a folder at all (411)

`s3_request_ext` set `Content-Length` only in the branch that HAS a body, the one branch that does not need it. A bodyless PUT went out with no `Content-Length`, and AWS S3, Backblaze, Cloudflare R2, Alibaba OSS, Storj, Tencent, Oracle and MEGA S3/S4 all answer `411 Length Required`. Two real operations go out that way: `mkdir` (the zero-byte directory marker) and `UploadPartCopy` (server-side copy past the 5 GiB single-PUT limit). `server_side_copy_single` hand-rolls its request and already set the header, which is why plain CopyObject was fine and the gap stayed invisible: the lab profile is MinIO, which tolerates the omission.

Re-running the twelve affected S3 profiles with the fix took **ten of them from broken to fully passing**.

### 4. FileLu's WebDAV files were undownloadable (206 treated as an error)

A plain GET, no Range sent, comes back `206 Partial Content` with `content-range: bytes 0-65535/65536`: the complete object. Both download paths matched only `StatusCode::OK`, so it failed three times and left no local file. The new check accepts a 206 only when Content-Range starts at 0 and reaches the last byte, so a genuinely partial body is still rejected rather than committed as complete.

### 5. FileLu's S3 gateway returned double-encoded names (LIVE-1 shape, new cause)

`sp ace & ünïcodé.txt` was listed as `sp ace & Ã¼nÃ¯codÃ©.txt`. The object is stored correctly, `stat` and `get` on the real name both succeed, so this is the LIVE-1 consequence again: a GET of the displayed name 404s and `sync --delete` reads the user's real local file as an orphan. Repaired for that endpoint only, and only when the key is entirely Latin-1 range and carries a mojibake lead byte, so correct keys are never rewritten.

### 6. Azure `rmdir` said "Removed empty directory" and left it there

`mkdir` writes a `<path>/` marker blob that `list` cannot see (queried with prefix `<path>/`, the marker's name IS the prefix, so it strips to an empty string and is dropped). `rmdir_recursive` only walked what `list` returned, so the marker survived every time. Found while cleaning up after the sweep.

### Findings that are NOT ours, recorded rather than patched

- **ImageKit**: upload returns 200 with a valid `fileId`, `filePath` and CDN URL, the object is genuinely served (`curl` returns 200 and the exact byte count), but the Media Library List API never returns it, with `path`, without `path`, with `searchQuery`, or with no filter at all. Our query is not the problem. Consequence is still severe: `sync --delete` against that remote plans `delete_local` for files that ARE stored.
- **Google S3 interop**: mkdir refused with 403, a bucket permission, not the 411 above.
- **Blomp (Swift)**: a listing that returned `[]` right after four successful uploads, and a bulk-delete that reported a failure on a percent-encoded container path. Not reproduced a second time, so it is recorded, not claimed.

### Verification after all six fixes

The eight most affected profiles were re-run end to end and all pass: AWS, Backblaze, Cloudflare R2, Azure Blob, FileLu S3, FileLu WebDAV, MinIO, Nextcloud. Every remote test directory created during the sweep has been removed and the removal verified.

Gates on the final head: `cargo test --lib providers::` 822 passed, 0 failed, 1 ignored; `cargo clippy --all-targets --features aerorsync -- -D warnings` clean, also under `--cfg ci_lane3`; `cargo fmt --all --check` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Azure Blob listings for nested paths and paginated results.
  * Recursive directory deletion now removes directory markers reliably.
  * Improved handling of STS error messages containing special characters.
  * Corrected filename display for affected FileLu S3 listings.
  * Improved compatibility for directory creation and multipart copy operations.
  * Strengthened WebDAV downloads to reject incomplete or invalid partial responses.

### DIFF
--- a/src-tauri/src/providers/azure.rs
+++ b/src-tauri/src/providers/azure.rs
@@ -450,7 +450,16 @@ impl AzureProvider {
 
     /// Parse XML blob list response using quick-xml event-based parser.
     /// Returns (items, next_marker) where next_marker is Some if pagination continues.
-    fn parse_blob_list(&self, xml: &str) -> (Vec<BlobItem>, Option<String>) {
+    /// Parse one page of a `comp=list` response.
+    ///
+    /// `strip_prefix` MUST be the blob prefix the request was issued with (the
+    /// `&prefix=` query value), not the provider's `current_prefix`: Azure
+    /// echoes every `<Name>` as a full path from the container root, so the
+    /// listed prefix is what makes an entry relative. Using `current_prefix`
+    /// instead made every one-shot `ls /a/b` (CLI, MCP, first GUI call, sync
+    /// scan) return an empty directory, because each blob still carried a `/`
+    /// and was dropped by the depth filter below.
+    fn parse_blob_list(xml: &str, strip_prefix: &str) -> (Vec<BlobItem>, Option<String>) {
         let mut items = Vec::new();
         let mut next_marker: Option<String> = None;
 
@@ -570,7 +579,7 @@ impl AzureProvider {
                     b"BlobPrefix" if in_prefix => {
                         let display_name = current_name.trim_end_matches('/');
                         let relative = display_name
-                            .strip_prefix(&self.current_prefix)
+                            .strip_prefix(strip_prefix)
                             .unwrap_or(display_name);
                         let relative = relative.trim_start_matches('/');
                         if !relative.is_empty() {
@@ -586,7 +595,7 @@ impl AzureProvider {
                     }
                     b"Blob" if in_blob => {
                         let relative = current_name
-                            .strip_prefix(&self.current_prefix)
+                            .strip_prefix(strip_prefix)
                             .unwrap_or(&current_name);
                         let relative = relative.trim_start_matches('/');
                         if !relative.is_empty() && !relative.contains('/') {
@@ -649,7 +658,11 @@ impl AzureProvider {
     /// Execute a paginated blob list request, returning all items across pages.
     /// AZ-004: Checks HTTP status before attempting XML parsing.
     /// AZ-005: Uses retry logic for transient errors.
-    async fn list_blobs_paginated(&self, base_url: &str) -> Result<Vec<BlobItem>, ProviderError> {
+    async fn list_blobs_paginated(
+        &self,
+        base_url: &str,
+        strip_prefix: &str,
+    ) -> Result<Vec<BlobItem>, ProviderError> {
         let mut all_items = Vec::new();
         let mut marker: Option<String> = None;
 
@@ -690,7 +703,7 @@ impl AzureProvider {
                 .await
                 .map_err(|e| ProviderError::ParseError(e.to_string()))?;
 
-            let (items, next_marker) = self.parse_blob_list(&body);
+            let (items, next_marker) = Self::parse_blob_list(&body, strip_prefix);
             all_items.extend(items);
 
             match next_marker {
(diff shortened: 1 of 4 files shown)


axpnet:
@coderabbitai full review

PR #540, opened 2026-07-31
fix(gui): make the delete confirmation reachable, and give overlays one stacking scale

### LINKED ISSUE #537: [BUG] Can't confirm deleting duplicates
**Describe the bug**
I encountered a bug in the Find Duplicates in AeroFile that at first I thought was an infinite loop, but now I see is just a faulty z-index or a similar nature.

**Steps To Reproduce**
1. Run Find Duplicates on a folder that has duplicates.

2. Click on the red '🗑 Delete Selected' when some duplicate files are ticked. Notice that the red '🗑 Delete Selected' button loads up forever.

3. Drag this '🔍 Find Duplicates' window horizontally all the way to either side, left or right, doesn't matter which.
	You'll see the popup that's supposed to confirm the deletion hidden behind. Trying to click its red Delete button does nothing.

4. Open File Explorer or whatever other file manager app you have in two tabs or windows: the Recycle Bin 🗑, and the folder where these duplicates are. Take note that you see the duplicates in their original folder and not in the Recycle Bin yet.

5. Click on the ✖ of the '🔍 Find Duplicates' window and then on the red Delete in the confirmation popup. Notice that now all the ticked duplicates moved from the original folder to the Recycle Bin. Thank you for making Find Duplicates only soft-delete, by the way.

**Expected behavior**
Be able to delete duplicates without needing to close the '🔍 Find Duplicates' window to confirm.

**Environment:**
 - OS: Windows 10
 - AeroFTP Version: v4.1.6

**Additional context**
An alternative way to confirm important decisions in GUI besides using a popup with its own buttons, is simply having to hold the original confirmation button for a few seconds. Just saying. This could be an alternative to try or think about.
For now, increasing the z-index of the confirmation popup or whatever is currently wrong there should be sufficient.

axpnet:
Fixes #537.

## What happened

Find Duplicates asked for confirmation twice. The dialog raises its own prompt, and the App callback behind `onDeleteFiles` raised a second one through `setConfirmDialog` whenever "confirm before delete" was on. `ConfirmDialog` is mounted ~300 lines earlier in `App.tsx` than the finder, and both sat at `z-50` — so at an equal z-index DOM order decided it. The finder painted over the very question it was waiting on, and its full-viewport backdrop swallowed every click aimed at the buttons underneath.

That accounts for each step in the report: the button "loads up forever" because the promise the callback awaited never settled; dragging the finder aside reveals the confirm but does not free it, because the panel moves and the overlay does not; and closing the finder unmounts the overlay, at which point the confirm becomes clickable and the delete goes through to the Recycle Bin.

## The fix

The dialog owns the confirmation outright. It takes `confirmBeforeDelete` and skips its own prompt when the setting is off; the callback only deletes. One question, in the window that asked it.

The stacking half is fixed for the whole app rather than for this one pair. `src/utils/modalLayers.ts` holds the scale — modal, a modal's own confirm, the elevated dialog tier, the app-wide confirm, the quit guard, the lock screens — and the sweep that came with it found two more of the same defect:

- The app-wide confirm had to clear more than `z-50`. Of the twelve `setConfirmDialog` call sites, the overwrite and delete prompts fire from flows started inside the z-9998..10000 tier (AeroSync, the transfer plan, the trash managers), so a confirm below them would have repeated this bug there.
- The lock screens sat at z-100 and z-200, under that same tier. `isAppLocked` gates no rendering in `App.tsx` — the whole tree stays mounted behind the lock overlay — so a dialog left open when the idle probe fired stayed legible on top of the lock screen. Both now sit above everything.

One more thing surfaced on the way: the in-dialog confirm asked for `duplicates.confirmDelete`, a key present in none of the 47 locale files, so `t()` printed the key name at the user instead of a question. It now uses `duplicates.deleteConfirm`, which exists.

## Pins

Four, each verified failing on the code it replaces:

- `modalLayerSweep.test.ts` walks all 144 `fixed inset-0` overlays in the source and refuses any that can cover the app-wide confirm or the lock screens, and any with no z-index at all.
- The `DuplicateFinderDialog` mount site in `App.tsx` must carry no `setConfirmDialog`.
- The dialog must branch on `confirmBeforeDelete` rather than confirm unconditionally.
- Every `t()` key the dialog uses must resolve in `en.json`, the fallback all 47 locales fall through to.

## On the alternative in the report

Hold-to-confirm is a real pattern, but not as a replacement for this one: a press-and-hold has no keyboard equivalent and is hard to complete with a motor impairment. Worth considering later as something a user can opt into, not as the default path to a delete.

## Summary by CodeRabbit

* **New Features**
  * Duplicate-file deletion now supports configurable confirmation prompts, preventing duplicate confirmations.
  * Improved layering ensures lock screens, confirmations, and dialogs appear in the correct order.

* **Bug Fixes**
  * Fixed modal stacking behavior across lock screens, guarded close dialogs, and duplicate-file dialogs.

* **Tests**
  * Added coverage for deletion confirmation behavior, translation keys, and modal layering consistency.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15982,18 +15982,13 @@ const App: React.FC = () => {
             isOpen={true}
             scanPath={duplicateFinderPath}
             onClose={() => setDuplicateFinderPath(null)}
+            // The setting governs the dialog's own confirmation. Raising a second
+            // one here is what made #537: `ConfirmDialog` is mounted far above in
+            // this tree, so it rendered behind the finder's overlay, unreachable
+            // behind its backdrop, while this promise never settled and the
+            // "Delete Selected" button span forever.
+            confirmBeforeDelete={confirmBeforeDelete}
             onDeleteFiles={async (paths) => {
-              // Respect confirmBeforeDelete setting
-              if (confirmBeforeDelete) {
-                const confirmed = await new Promise<boolean>(resolve => {
-                  setConfirmDialog({
-                    message: t('duplicates.deleteConfirm', { count: paths.length }),
-                    onConfirm: () => { setConfirmDialog(null); resolve(true); },
-                    onCancel: () => { setConfirmDialog(null); resolve(false); },
-                  });
-                });
-                if (!confirmed) return;
-              }
               for (const p of paths) {
                 try {
                   await invoke('delete_to_trash', { path: p });
--- a/src/components/AccountLockScreen.tsx
+++ b/src/components/AccountLockScreen.tsx
@@ -24,6 +24,7 @@ import {
 } from '../utils/userPartitions';
 import { PROFILES_CHANGED_EVENT } from '../utils/serverProfileStore';
 import { mapUserPartitionError } from '../utils/userPartitionErrors';
+import { MODAL_Z } from '../utils/modalLayers';
 
 const LOCK_PATTERN_KEY = 'aeroftp_lock_pattern';
 const DEFAULT_PATTERN = 'isometric';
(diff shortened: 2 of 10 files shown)


axpnet:
@coderabbitai full review

axpnet:
Valid, and the more interesting half is the one the finding does not spell out. Fixed in 50cc6bd.

Both writes are now functional updates keyed by the set of paths actually deleted, so a newer scan survives and only what was deleted gets unticked, rather than the whole selection being cleared.

But reconciling correctly only makes the outcome sane; it does not stop the race. The three controls that start a scan (the two mode toggles and the fuzzy-cutoff field) were live throughout a delete, so a scan could complete against a tree the delete had already changed. Those are now disabled while `isDeleting`, which is what actually closes it. Tidying up after a race and preventing it are different fixes and this needed both.

Pinned in `DuplicateFinderDialog.test.ts`: the callback may not close over `groups`, both writes must be functional, and all three scan-starting controls must carry the guard. Verified failing on the previous commit.

PR #541, opened 2026-07-31
i18n: add the 32 keys the UI asked for and never had

axpnet:
Stacks on #540 — please merge that one first.

## What the sweep found

#540 started from one string: the Find Duplicates confirmation asked for `t('duplicates.confirmDelete')`, a key that existed in none of the 47 locale files, so `t()` did what it does on a miss and printed the key name at the user instead of a question. Asking the same question of the whole source turned up 31 more, and every one of them is reachable — a column header in the B2 hidden-file manager, all four labels in the GitHub branch selector, the version browser's entire vocabulary, the "Later" on an update prompt.

The `t('x') || 'Fallback'` idiom at most of those call sites was not protecting anything: `t()` returns the key, the key is truthy, so the fallback was dead code. One of them had been carrying the Italian string `'Più tardi'` as the English fallback — which is a fair measure of how long nobody had read that line.

## What this does

Fifteen of the keys name a string the app already had, translated and reviewed, under a different key: `common.path` beside `browser.path`, `errors.unknown` beside `transfer.errorKind.unknown`, `toast.transferFailed` beside `aeroShare.receive.failedTitle`. Those take the existing translation from each of the 47 locales — no new wording, nothing invented.

The remaining seventeen are genuinely new strings, and they are translated in all 46 locales. `npm run i18n:validate` reports 46/46 clean: no missing keys, no orphans, no `[NEEDS TRANSLATION]` placeholders. The dead `|| '...'` fallbacks are removed from the 13 call sites that carried them.

## The pin

`src/i18n/noPhantomKeys.test.ts` asks it of the whole source from here on: every literal `t('a.b')` must resolve in `en.json` — the file every other locale falls back to before `t()` gives up and returns the key — and must resolve to a string rather than to a branch of the tree, which would render as `[object Object]`.

Verified failing on the tree it replaces, listing all 32.

## Summary by CodeRabbit

* **New Features**
  * Improved Duplicate Finder deletion flow with confirmation handling, selection preservation, refreshed results, and disabled controls during deletion.
  * Added consistent modal layering for lock screens, confirmations, and other overlays.
* **Localization**
  * Expanded translated labels and messages across many languages, covering transfers, cloud storage, GitHub, privacy, errors, renaming, and settings.
  * Removed English fallback text so supported interfaces consistently use translations.
* **Bug Fixes**
  * Improved overlay stacking and asynchronous deletion behavior.
* **Tests**
  * Added coverage for translation keys, modal layering, and Duplicate Finder workflows.


axpnet:
@coderabbitai full review

axpnet:
Four of the five applied in c81119d (on the #540 branch this stacks on, since those files originate there). The fifth is declined with a reason.

**Partial delete failure.** Valid and now fixed: `onDeleteFiles` walks the list, so a reject can arrive with earlier paths already gone. Which ones is not knowable from the dialog, so it re-scans, which is the only answer that cannot be wrong.

**The z-token window.** Valid, and stricter parsing paid for itself immediately. Reading the `className` value that actually contains the match (quoted string or `{...}` expression) instead of a five-line window turned up a false positive worth keeping out: `SaveAllMenu` has a comment explaining why its confirm is portalled, and the words `fixed inset-0` inside it were being counted as an overlay with no z-index. Comments are now blanked before the scan, newlines preserved so line numbers still point at the right place.

**Module-scope `expect` and `readZ`.** Both valid. Collection is lazy inside the tests now, the "every overlay has a z-index" check is a named test rather than a bare `expect` during evaluation, `modalZIndexOf` is imported instead of reimplemented, and the eslint-disable that had no `any` to suppress is gone.

**Elevated-tier ordering.** Added, plus one the suggestion did not ask for: no two layers may share a number. An accidental tie is precisely what #537 was.

**`fi.json` SSE wording.** Declined, and it is worth being specific because the finding reads only the first sentence. The English source already carries the caveat verbatim: *"An SSE-KMS or SSE-C ETag can still be 32 hex characters and then passes the same shape check as a real digest, so a reported ETag is advisory for such an object."* The Finnish is a faithful translation of that, including the closing clause. There is nothing to correct, and the line is outside this PR in any case, which touched `fi.json` only to add keys.

**Behavioral assertions over source text.** Same answer as on #543: the repo has no DOM harness (vitest on `environment: node`, no jsdom, no testing-library), and the `?raw` source scan is the established convention here. Adding that harness is a dependency and lockfile change that belongs in its own PR, not smuggled into this one.

PR #542, opened 2026-07-31
fix(gui): show file thumbnails whole instead of centre-cropped

axpnet:
From the 2026-07-31 wishlist batch on #347:

> **Where:** AeroFile 📁 in Icons and Large Icons view. **What:** Please don't hide the edges of images. Each icon should show its image in full.

Right, and for a reason worth naming: `object-fit: cover` fills the square by cropping whatever does not fit. On a wide photo that is both edges; on a screenshot it is usually the part that would have told you which screenshot it is. A thumbnail exists to identify the file, so cropping the identifying part out of it defeats the view.

Three components render a file thumbnail — `ImageThumbnail` (the Icons grid, and the remote panel), the Large Icons tile, and the provider-supplied one — and all three now state `object-contain` themselves.

`.file-grid-thumbnail` keeps the box: the 48px square, the radius, and the background the letterboxing sits on. It no longer decides the fit. That part is deliberate rather than tidy-minded — a `?raw` import of a stylesheet comes back empty under Vite, so a rule left in `styles.css` would have been the one part of this decision no test could read. With the fit stated in the components, all three surfaces are checkable and none of them can contend with the stylesheet.

The boxes keep their size, so nothing in the grid reflows. Avatars and the chat attachment chips keep `cover`: those are decorative crops of a known shape, not a file listing.

Pinned by `fileThumbnailFit.test.ts`, verified failing on `main` across all five assertions.

### DIFF
--- a/src/components/ImageThumbnail.tsx
+++ b/src/components/ImageThumbnail.tsx
@@ -52,7 +52,10 @@ export const ImageThumbnail: React.FC<ImageThumbnailProps> = ({
     if (error || !src) {
         return <div className="file-grid-icon">{fallbackIcon}</div>;
     }
-    return <img src={src} alt={name} className={className || "file-grid-thumbnail"} />;
+    // `object-contain`: show the file whole. `cover` crops whatever does not fit
+    // the square, which on a wide photo is both edges and on a screenshot is
+    // usually the part that identifies it (discussion #347).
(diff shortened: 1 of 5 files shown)


axpnet:
@coderabbitai full review

axpnet:
@coderabbitai full review

PR #549, opened 2026-08-01
fix(mcp): guard destructive defaults in aeroftp_speed and remote_trash empty

axpnet:
Closes audit findings MCP-1 and MCP-2 (pre-tag audit, docs/security-evidence/SECURITY-EVIDENCE-v4.1.7.md §8.3).

MCP-1: `aeroftp_speed` and `cli speed --remote-path` uploaded a test payload to any caller-named remote path and deleted it afterwards, with no existence check: naming a real user file overwrote and then deleted it. Both surfaces now stat the remote path after connecting and refuse if it already exists (the error suggests omitting `remote_path` for an auto-generated scratch path). No overwrite flag added.

MCP-2: `remote_trash empty` defaulted `dry_run=false` with an empty prefix, so one call irreversibly purged every version and delete marker in the bucket. It now defaults `dry_run=true`, and a real whole-bucket purge additionally requires `confirm_whole_bucket: true`; a scoped prefix purges without the flag. Response shape unchanged.

Also fixed (same class, found during verification): CLI `versions empty-trash` skipped the TTY confirmation in non-interactive runs and purged silently; it now refuses without `--yes` when non-interactive.

Tests: 7 new unit tests (existing-path refusal, fresh scratch path ok, dry-run default, whole-bucket confirm gate, scoped purge, consent helper). Gates: cargo test (ai_core + cli) green, clippy -D warnings clean incl. `--cfg ci_lane3`, fmt clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Empty-trash operations now default to a safe dry run.
  * Whole-bucket purges require explicit confirmation, including `--yes` for non-interactive CLI use.
  * Scoped purges remain available without whole-bucket confirmation.
  * Remote speed tests refuse existing paths and fail safely when caller-specified paths cannot be verified.
  * Missing paths are distinguished from authorization and transient errors during cleanup.
* **Tests**
  * Added coverage for purge confirmation, dry-run behavior, path protection, and cleanup safeguards.

### DIFF
--- a/src-tauri/src/ai_core/remote_tools.rs
+++ b/src-tauri/src/ai_core/remote_tools.rs
@@ -1251,8 +1251,9 @@ async fn remote_versions(ctx: &dyn ToolCtx, args: &Value) -> Result<Value, ToolE
 /// include_noncurrent to include older versions); undelete drops a delete marker
 /// so the object reappears; restore copies an older version forward; purge
 /// permanently removes one version or marker; empty purges the whole trash under
-/// the prefix (dry_run previews the count/bytes without deleting). All destructive
-/// actions are irreversible.
+/// the prefix (defaults to dry_run=true, a preview that deletes nothing; a real
+/// whole-bucket purge with an empty prefix also needs confirm_whole_bucket=true).
+/// All destructive actions are irreversible.
 async fn remote_trash(ctx: &dyn ToolCtx, args: &Value) -> Result<Value, ToolError> {
     let server = normalize_server(args)?;
     let action = get_str_opt(args, "action").unwrap_or_else(|| "list".to_string());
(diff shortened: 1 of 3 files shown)


axpnet:
@coderabbitai review

axpnet:
Follow-up pushed in b98a7c613, addressing all three review findings.

> `backend.stat(&remote_path).await.is_ok()` treats every error as missing

Correct, and it was the same fail-open shape this PR set out to close: an expired token, a dropped connection, a provider 5xx or an unsupported `stat` all passed the guard and handed the upload-then-delete sequence a path that may hold user data. Both surfaces now proceed only on an explicit provider not-found and refuse with the underlying error otherwise.

The auto-generated UUID scratch path stays exempt by design: it holds no user data by construction, so a flaky or unsupported `stat` must not cost a legitimate speed test. The strict check applies to the path the caller named, which is the one that can be anything.

On the atomic no-clobber suggestion: adding an exclusive-create primitive to `RemoteBackend` would have to be implemented across every provider, and the ones that cannot offer the guarantee (plain FTP among them) would still fall back to lookup-then-write. It is a real improvement but a separate change with its own blast radius, not a follow-up to a safety fix. Failing closed removes the destructive outcome here; the residual race is a second caller creating the caller-named path inside the check-to-upload window, which is bounded and no longer silent.

> The help text says that `--yes` is required for every non-interactive run

Fixed — it now says "required for real non-interactive purges".

The CLI decision moved into `ensure_speed_path_is_free`, a pure helper beside `ensure_empty_trash_consent`, so it is covered by a test instead of only being reachable with a live provider. Three new tests, each verified to fail against the previous guard and pass against this one. Gates: `cargo test --lib` 3442 passed, fmt clean, `clippy --all-targets --features aerorsync -D warnings` clean.

PR #622, opened 2026-08-25
fix(intro-hub): CUSTOM / GENERIC SERVERS gets the shared badge order and its own method glyphs

axpnet:
Closes both halves of the CUSTOM / GENERIC SERVERS report on #347: the ordering follow-up ([17978401](https://github.com/axpdev-lab/aeroftp/discussions/347#discussioncomment-17978401)) and the shared-symbol complaint ([17859082](https://github.com/axpdev-lab/aeroftp/discussions/347#discussioncomment-17859082)). Both replies had been sitting unanswered since 11 August.

## 1. The order

v4.1.7 unified the Add Service badges with the Quick Connect tabs on one presentation order, `webdav, sftp, ftps, ftp, s3`, exported as `PROTOCOL_BADGE_ORDER` precisely so the two could not drift apart again. Ehud replied: "Thanks, but the CUSTOM / GENERIC SERVERS are still not in that order."

He is right, and it is the same defect rather than a leftover of it. `CUSTOM_PROFILES` never consulted the rank; it rendered in the order the array happened to be authored, `FTP, SFTP, S3, WebDAV`. A third hand-written opinion of the sequence, one screen below the two that had just been made to agree.

It sorts by `protocolBadgeRank` now, rather than being reordered by hand. Reordering the literal fixes today's display and leaves the next entry free to land anywhere; sorting means an entry appended at the bottom still appears in the right place.

## 2. The icons

"All protocols share the same generic symbol in CUSTOM / GENERIC SERVERS at the bottom of the table view of the Add Service page."

Also right: the strip drew one `Server` icon for all four entries, so FTP, SFTP, S3 and WebDAV were visually identical in the one place whose whole job is telling them apart.

Nothing new was needed. `ConnectionMethod`, `CONNECTION_METHOD_GLYPH` and `methodIcon` already exist on `main` and already serve the badges elsewhere; the strip simply never used them. Each entry now carries its method and draws that method's glyph, so the symbols stay in agreement with the rest of the UI by construction instead of being chosen again here.

## The tests

`customProfilesOrder.test.ts` pins the order against the shared rank, spells out the expected sequence so a change to the shared order has to be deliberate here too, checks every entry is ranked (an unranked protocol scores -1 and would jump to the front as if it were a native API), and pins the glyph pairing rather than only the presence of a glyph, since a wrong symbol is worse than the generic one it replaces.

Every assertion was run against the defective code first. The ordering ones fail with exactly the order reported:

```
AssertionError: expected [ 'ftp', 'sftp', 's3', 'webdav' ] to deeply equal [ 'webdav', 'sftp', 'ftp', 's3' ]
```

and collapsing every entry onto one method fails both glyph assertions:

```
AssertionError: expected 1 to be greater than 1
AssertionError: expected [ [ 'webdav', 'FTP' ], …(3) ] to deeply equal [ [ 'webdav', 'WebDAV' ], …(3) ]
```

## On the overlap with `fix/introhub-347`

That parked branch (`c4897aa7d`, no PR, 11 days idle, 10 commits behind `main`) changes the same block, and an earlier comment here proposed resolving that at merge time. Folding the icons in instead is the better answer, so that comment is superseded: with both halves here the two no longer disagree about this block at all, and the resolution when that branch is revived is simply to drop its `CUSTOM_PROFILES` hunk.

What that branch keeps is the work that is genuinely its own and is not touched here: the Add Service form-tab reorder, and the icon changes in `IntroHubHeader`, `CatalogTable`, `discoverData` and the shared glyph map itself.

## Verification

`tsc --noEmit` clean. Full `vitest run` green: 86 files, 753 tests.

### DIFF
--- a/src/components/IntroHub/DiscoverPanel.tsx
+++ b/src/components/IntroHub/DiscoverPanel.tsx
@@ -19,7 +19,7 @@ import { useDiscoverHealthCheck } from '../../hooks/useDiscoverHealthCheck';
 import { openUrl } from '../../utils/openUrl';
 import { middleClickOpen } from '../../utils/middleClick';
 import { CatalogTable, loadTierFilter, persistTierFilter, TIER_FILTERS, type TierFilter } from './CatalogTable';
-import { PROVIDER_CATALOG, companyInCategory, companyTierInCategory, isDevOnlyProvider } from '../providerCatalog';
+import { PROVIDER_CATALOG, companyInCategory, companyTierInCategory, isDevOnlyProvider, protocolBadgeRank } from '../providerCatalog';
 
 /** All category sidebar entries share these keys; 'all' is a virtual category. */
 type DiscoverCategoryId = CatalogCategoryId | 'all';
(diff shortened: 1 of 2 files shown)


axpnet:
## Overlap with `fix/introhub-347`, declared rather than discovered at merge time

`fix/introhub-347` (`c4897aa7d`, parked, no PR, 11 days idle, 10 commits behind `main`) touches this exact block. It is not a duplicate of this PR but it is not disjoint either, so the resolution should be a decision rather than whatever git produces:

- **It reorders the `CUSTOM_PROFILES` literal by hand** to `WebDAV, SFTP, FTP, S3`. Same visible result as this PR today. The difference is what happens to the next entry someone appends: a hand-ordered literal lets it land anywhere, a sorted one does not. There is also no test there.
- **It gives each entry a `method` glyph** from a shared map, replacing the single `Server` icon used for all four. That is the icons half of the report ([17859082](https://github.com/axpdev-lab/aeroftp/discussions/347#discussioncomment-17859082)) and it is genuinely that branch's work. Nothing here touches it.

So the two are complementary, and the intended shape when that branch lands is: keep its `method` field and its glyph rendering, keep the `.sort(...)` and `customProfilesOrder.test.ts` from here, and drop its hand-ordering of the literal since the sort makes it redundant. The test in this PR then covers its ordering too.

Merging this first is deliberate: the report has been unanswered since 11 August and the ordering fix is small, tested and independent of the icon work, which is the larger and slower half.

PR #630, opened 2026-08-26
fix(transfer): do not evict a concurrent opener's checkpoint record

axpnet:
## Why

Tracker #628 item 15 (R-01). Concurrent `enforce_cap` calls at a full store of same-second non-terminal records could evict each other's just-written opening slot: the sort was terminal-first then oldest, and equal timestamps fell through to readdir order.

A five-second grace window reorders the pick so an older occupant (and terminal residue) is preferred over a sibling still in flight.

The first cut on this branch skipped those fresh records and **failed the new open** when every occupant was still inside the window. A DAG fan-out that opens more than `DEFAULT_CHECKPOINT_MAX_RECORDS` (256) multipart sessions in five seconds would have started failing at the 257th. That trades a theoretical race for a practical failure under load, so it is not the behaviour here.

Fresh records now sort last rather than being skipped. The cap still holds: the 257th open succeeds by dropping the oldest occupant, which is the documented cost of a full store of resumable records. `open_or_create` is not refused because of the grace window.

Stacked on #629.

## Test plan

- [x] 20 `transfer_dag::checkpoint` unit tests, including: prefer an old record over a fresh sibling; still evict fresh terminal residue; still accept the next open when every occupant is in-grace
- [x] `cargo fmt --all`
- [x] `cargo clippy --all-features --tests -- -D warnings`
- [x] `cargo clippy --bin aeroftp-cli --all-features --tests -- -D warnings`
- [x] `cargo clippy --all-targets -- -D warnings` (build.yml:191 default-features lane)

Tracker: https://github.com/axpdev-lab/aeroftp/issues/628

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved checkpoint retention when storage capacity is reached.
  - Recently updated resumable checkpoints now receive a five-second grace period before eviction.
  - Eviction prioritizes completed records and older resumable records, while still enforcing the maximum capacity when necessary.

- **Tests**
  - Added coverage for eviction ordering, grace-period behavior, and capacity-limit scenarios.

### DIFF
--- a/src-tauri/src/transfer_dag/checkpoint.rs
+++ b/src-tauri/src/transfer_dag/checkpoint.rs
@@ -27,7 +27,22 @@ pub const DEFAULT_CHECKPOINT_TTL: Duration = Duration::from_secs(7 * 24 * 60 * 6
 /// resumable transfers. The documented cost is that the oldest resumable
 /// transfer can be dropped once the store is full of resumable records; the
 /// explicit escape for a decommissioned endpoint is `forget_endpoint`.
+///
+/// A just-opened resumable record is not evicted for
+/// `DEFAULT_CHECKPOINT_EVICT_GRACE`, so two concurrent `open_or_create` calls
+/// at a full store cannot delete each other's opening record. Terminal
+/// residue is always eligible: a commit that just finished is the thing the
+/// cap exists to reclaim.
 pub const DEFAULT_CHECKPOINT_MAX_RECORDS: usize = 256;
+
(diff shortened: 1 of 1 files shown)


axpnet:
@coderabbitai review

axpnet:
Superseded by #633. GitHub closed this automatically when the branch it was stacked on was deleted on merge, and it cannot be reopened once its base is gone. The commits are unchanged, rebased onto `main`.

PR #631, opened 2026-08-26
feat(swift): reach a private OpenStack, and give the Blomp tile its identity

axpnet:
Tracker [#628](https://github.com/axpdev-lab/aeroftp/issues/628) item 10, plus a live defect found while giving that item its control.

Stacked on [#630](https://github.com/axpdev-lab/aeroftp/pull/630), which is stacked on [#629](https://github.com/axpdev-lab/aeroftp/pull/629). Review the top commit only.

### The defect

Picking Blomp from the protocol grid and connecting **without saving first** was refused with the cleartext object-store error.

The Blomp tile is the only preset tile that did not declare `providerId`, so the protocol change took the plain-protocol path instead of the provider branch. The auth URL arrived from a normalizer fallback while the identity did not arrive at all, and `provider_id` reached the backend as `null`. The exemption that lets Blomp's HTTP object store through the v4.1.8 downgrade guard is keyed on exactly that id, so the guard refused. Saving the profile first worked, because both save paths inject the id, which is how this survived the release.

### The capability

Swift is a protocol and not only Blomp, but the form hardcoded Blomp's Keystone and rewrote it on every keystroke in the username field, so a private OpenStack could not be configured at all, cleartext or otherwise. The auth URL is editable now, and a deployment that legitimately publishes a cleartext object store accepts it for itself through a checkbox that appears only once the URL is no longer the preset's.

The identity follows the auth URL rather than the protocol, in both save paths and ahead of the stored id, so editing a Blomp profile into a private OpenStack cannot carry Blomp's exemption across, and editing it back restores it. Without that rule the new field would have handed every custom Keystone an exemption it never asked for.

### Notes for the reviewer

The refusal message names the control now that one exists, and names `allow_cleartext_storage_endpoint` beside it for the CLI, MCP and AeroCloud profiles that never see a form.

The assertion that the message must NOT name a control was inverted rather than deleted. It is worth a look: the new wording happened to keep passing it, because it avoided the two words the assertion forbade while naming a control anyway. That is the same shape as the tile that carried no id, and as the logo allowlist in #629.

### Verified green

`cargo fmt`, clippy at `--all-features --tests` on the lib and on the `aeroftp-cli` bin and at `--all-targets` on default features, the full Rust suite, `tsc`, vitest at 93 files and 801 tests, `i18n:validate` with 0 missing and 0 placeholders across 47 locales.

Two new keys translated into all 46 non-English locales.

https://claude.ai/code/session_017gpkDULZYjdzke3wULZyDw


axpnet:
@coderabbitai review

PR #643, opened 2026-08-27
feat(aerorsync): close integrated protocol and metadata gaps

axpnet:
## Summary

- add protocol-31 ACL negotiation, wire codec, Linux fd-bound capture/apply, and production opt-in
- fail closed on unsupported checksum winners and metadata-flag mismatches before file-list commit
- align product eligibility and metadata flags across platforms without enabling owner/group/device transfer
- add stock-rsync ACL upload/download coverage, exact product-path ACL+xattr coverage, and Linux package runtime-contract gates

## Scope

This closes the integrated AeroRsync gaps first. It does not create or publish the standalone `aerorsync` crate; that planning begins only after this branch is integrated and stable.

## Verification completed locally

- `cargo fmt --all -- --check`
- feature-on and feature-off `cargo check`
- strict Clippy (`-D warnings`)
- AeroRsync suite: 673 passed, 0 failed, 15 ignored
- exact product-path live ACL+xattr test against the key-auth fixture: 1 passed, 0 skipped
- all 15 stock-rsync 3.2.7 lane-3 tests: 15 passed, 0 skipped
- workflow YAML / JSON / shell syntax checks

Linux bundle assertions are intentionally validated by the Build workflow against the actual generated deb/rpm/AppImage and Snap artifacts.

## Summary by CodeRabbit

* **New Features**
  * Added Linux POSIX access ACL preservation for native delta uploads and downloads, including named-user permissions.
  * Expanded native delta synchronization across supported SSH authentication methods and platforms.
  * Added ACL-aware packaging for Linux desktop and Snap distributions.

* **Bug Fixes**
  * Unsupported or incompatible checksum negotiations now fail clearly instead of silently skipping verification.
  * Improved handling and reporting of unsupported filesystem metadata capabilities.


axpnet:
Pushed `4f6182276` after rebasing cleanly onto current `main` (`e65f06033`).

The Linux red was a real gate/layout mismatch plus an RPM dependency gap:

- the AeroRsync consumer is the shipped CLI/MCP payload, not the GUI or dispatcher
- the `readelf` regex did not match the literal `(NEEDED)` output shape
- Tauri does not auto-generate RPM requirements for the CLI added through `rpm.files`, so `libacl.so.1()(64bit)` is now explicit
- deb, AppImage and Snap checks now inspect `usr/lib/aeroftp/aeroftp-cli`; Snap also verifies the bundled runtime library

Verified CodeRabbit findings fixed in the same commit: target-safe native libacl fallback, immediate errno capture after `acl_set_fd`, Linux-only ENOTSUP seam, short-option parsing only before `.`, exact product command contract, XMIT_SAME_MODE caveat, and documented counts 673/15. The positional-tuple refactor was left out as non-functional churn. The requested Windows gate had already passed on the prior PR head.

Local gates after the fix:

- fmt, feature-off/on check, strict Clippy: PASS
- AeroRsync suite: 673 passed, 0 failed, 15 ignored
- exact product ACL+xattr live: 1 passed, 0 skipped
- lane3 live: 15 passed, 0 skipped (serialized); the checksum case also passed isolated after one concurrent fixture banner refusal
- release `aeroftp-cli`: `DT_NEEDED libacl.so.1`

PR #649, opened 2026-08-28
refactor(ftp): converge both FTP listing parsers onto one shared module

axpnet:
## Why this is not a pure move

Two FTP implementations exist and each carried its own copy of the listing parser. `providers::ftp::FtpProvider` is the one most surfaces use; the legacy `crate::ftp::FtpManager` is still reached by the AI agent fallback, `gui_tools`, `cloud_service`, the transfer-queue scan and the file panel's no-protocol branch. The copies had drifted apart, so a defect fixed in one stayed open in the other, and every later fix in this campaign would have had to be written twice or would have quietly covered only half the product.

The plan was a pure move. It cannot be one: the copies differ in ten measurable ways, and no shared module can serve both without changing one of them. Preserving both behaviours would mean a parser whose API carries six flags whose only purpose is to keep a defect alive, and a later change would have to remove them again.

The point of separating a move from a fix is that a regression stays attributable. Purity was the means, not the goal, so the goal is met here with evidence instead: a characterisation battery of the same synthetic rows was frozen against **both** parsers before the change. The provider's baseline is byte-identical afterwards, which is the proof that side did not move. The legacy's differences are enumerated below, every one of them.

Nothing was fixed here. Each of the ten changes is the legacy adopting what the provider already did.

## The ten changes, row by row

| # | row | before | after | reaches consumers |
|---|---|---|---|---|
| 1 | `lrwxrwxrwx    1 user     group           7 Jan 20 10:00 link -> target` | name="link" path="/link -> target" dir=false size=Some(7) perms=Some("lrwxrwxrwx") mod=Some("Jan 20 10:00") | name="link" path="/link" dir=false size=Some(7) perms=Some("lrwxrwxrwx") mod=Some("Jan 20 10:00") | yes |
| 2 | `-rw-r--r--    1 user     group        ???? Jan 20 10:00 odd.txt` | name="odd.txt" path="/odd.txt" dir=false size=None perms=Some("-rw-r--r--") mod=Some("Jan 20 10:00") | name="odd.txt" path="/odd.txt" dir=false size=Some(0) perms=Some("-rw-r--r--") mod=Some("Jan 20 10:00") | yes |
| 3 | `-rw-r--r-- 1 user group 123 Jan 20 10:00` | name="group 123 Jan 20 10:00" path="/group 123 Jan 20 10:00" dir=false size=None perms=None mod=Some("-rw-r--r-- 1") | row rejected | yes |
| 4 | `drwxr-xr-x    2 user     group        4096 Jan 20 10:00 .` | name="." path="/." dir=true size=Some(4096) perms=Some("drwxr-xr-x") mod=Some("Jan 20 10:00") | row rejected | no, already filtered |
| 5 | `drwxr-xr-x    2 user     group        4096 Jan 20 10:00 ..` | name=".." path="/.." dir=true size=Some(4096) perms=Some("drwxr-xr-x") mod=Some("Jan 20 10:00") | row rejected | no, already filtered |
| 6 | `total 12` | name="total 12" path="/total 12" dir=true size=None perms=None mod=None | row rejected | no, already filtered |
| 7 | `01-23-24  10:30AM       <DIR>          folder` | name="folder" path="/folder" dir=true size=None perms=None mod=Some("01-23-24 10:30AM") | name="folder" path="/folder" dir=true size=Some(0) perms=None mod=Some("01-23-24 10:30AM") | yes |
| 8 | `01-23-24 10:30AM 12345 a b c d e f` | name="f" path="/f" dir=false size=None perms=Some("01-23-24") mod=Some("c d e") | name="f" path="/f" dir=false size=Some(0) perms=Some("01-23-24") mod=Some("c d e") | yes |
| 9 | `not-a-date 10:30AM <DIR> folder` | name="folder" path="/folder" dir=true size=None perms=None mod=Some("not-a-date 10:30AM") | row rejected | yes |
| 10 | `drwxr-xr-x 2 1001 1001 4096 Jul 21 09:41 .` | name="." path="/." dir=true size=Some(4096) perms=Some("drwxr-xr-x") mod=Some("Jul 21 09:41") | row rejected | no, already filtered |
**Four of the ten never reach a consumer.** The caller already filtered them before this change, at `src/ftp.rs`: rows beginning `total ` or `Total `, recursive-listing directory headers ending in `:`, and entries named `.` or `..` dropped by name after parsing. So the observable delta is six rows, not ten, and a reader should not infer a larger change than the one that happened. Those three guards are now dead code, which is exactly the pattern this campaign exists to remove: a compensation left standing after the responsibility moved down. They are not removed here, because removing them would be a fix inside a convergence; they belong to the next change.

**Three of the six visible ones are the parser inventing data.** Row 3, an eight-token Unix row, arrived as a file named after a fragment of the listing line. Row 9, a row with no DOS date, arrived as a valid directory. And row 6, an `ls` header, arrived as a directory that does not exist. The legacy parser had no way to fail: whatever it could not read fell through to a fallback that took the raw line as a filename and guessed `is_dir` from whether the name contained a dot. An entry that is missing makes an operation do nothing. A directory that does not exist makes it do something: a recursive delete descends into it, a sync recurses, a tree enumerates it. Row 6 is not reproducible on either lab server, which are vsftpd and emit no `total` line, so it is covered by a synthetic row rather than by a live test.

**Row 1 is the symlink path.** The name was split on the arrow but the path was built before the split, so `link -> target` stayed inside it and any operation using that path addressed a file whose name literally contained the arrow.

## The size rows, stated precisely

Rows 2, 7 and 8 lose a distinction: `size: None` becomes `Some(0)`. `RemoteEntry.size` is a plain `u64` and cannot say "unknown", while the legacy type could.

The provider does not change here, and the wording matters because it drives a queued decision: this is not "the legacy behaviour was better and we chose the worse one". It is the shared type having no way to express the unknown. Every one of the seven consumers of `RemoteFile.size` already collapses it with `.unwrap_or(0)` (`lib.rs` at four sites, `ai_core/tauri_impl.rs`, `cloud_service.rs`, `transfer_queue_scan.rs`), so no caller ever observed the difference. The loss is in the type, not yet in the behaviour.

Recovering the real value where a listing reports 0 for a non-empty file is a separate change and is queued: it moves the hydration that the CLI does today down into the provider, so all three surfaces get it, and it needs no type change. Giving callers a way to say "unknown" is a further step across every provider, and it is recorded as its own item so that a change touching hundreds of call sites is not justified by a regression that does not exist yet.

## One structural addition

`RemoteFile` gains a `link_target` field. It is not a behaviour change and so does not appear in the table above, which is precisely why it is called out here: it changes the shape serialised to the frontend, and a reader scanning a list of behaviour changes would not find it.

It is additive and no consumer reads it yet, so it cannot produce a regression. It is here rather than later because this side discarded symlink targets entirely, and the queued change that aligns the symlink contract across surfaces cannot be honoured by a surface that has no symlink data to honour it with.

## What the review checked, and what it did not

The cross-review compared the two legacy baselines rather than reading the claimed list, and then did the check that matters: it normalised the common fields and compared the converged legacy against the **provider's** baseline over the shared inputs. Zero residual divergences, so every change is accounted for by "the provider already did this" and none is an invented correction. It confirmed that a row that now fails is skipped by the caller's `if let Ok(...)` rather than aborting the listing, which was the real risk of giving the parser a way to fail.

It did not run the tests, and says so: it verified that the assertion exists and is frozen, not that it passes. It did not read the module line by line against the old code, relying on the baseline, which covers 28 rows and not every possible input. Of the legacy's call sites it examined `.size`, the handling of `.` and `..`, and the listing loop's error path, and nothing else.

One correction to its report, in the same direction: it counted three invisible rows. There are four. Row 10, the vsftpd row with numeric owner and group, was also named `.` and so was already dropped by the same filter.

## Gates

`cargo fmt --manifest-path src-tauri/Cargo.toml --all -- --check`: 0.
`cargo clippy --manifest-path src-tauri/Cargo.toml --lib --tests -- -D warnings`: 0.
Both characterisation batteries pass, and `providers::ftp` is 17 passed, 0 failed.
The full suite result is added below once it finishes; two runs were killed by the environment earlier today, so compiling and running are done as separate steps.

Tracker: #628

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved FTP directory listing compatibility across Unix, DOS, and MLSD/MLST formats.
  - Symlink destinations are now preserved and displayed correctly.
  - Invalid, summary, and navigation entries are no longer shown as files or folders.
  - Unknown file sizes are handled consistently without disrupting listings.

- **Reliability**
  - FTP listings now provide more accurate file types, paths, timestamps, and metadata.

### DIFF
--- a/src-tauri/src/ftp.rs
+++ b/src-tauri/src/ftp.rs
@@ -42,6 +42,11 @@ pub struct RemoteFile {
     pub is_dir: bool,
     pub modified: Option<String>,
     pub permissions: Option<String>,
+    /// Where a symlink points. This side used to discard it and leave the
+    /// `name -> target` arrow inside `path`; PR-D needs it to align the
+    /// symlink contract across surfaces, and a surface with no data cannot
+    /// honour a contract.
+    pub link_target: Option<String>,
 }
 
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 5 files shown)


axpnet:
@coderabbitai review

This pull request was merged without a review: the only bot comment on it is a rate-limit notice from 20:15Z, and the merge happened at 00:03Z, so the quota had long since regenerated but the review was never retried. Requesting one now on the merged diff, since the change converges two FTP listing parsers onto one shared module and its behavioural deltas are the thing worth a second pair of eyes.

PR #651, opened 2026-08-29
fix(ftp): answer the questions the server was already answering

axpnet:
## One shape, six times

Every defect here has the same form: the server answered, the answer was discarded, and a later surface compensated for the loss. Each compensation lived in the CLI, so each was right on one surface out of three while the GUI and MCP kept the raw defect.

Two of the six turned out to be a different thing, and the distinction only appeared when a real server contradicted a premise written confidently from the documentation. **Four were data the server gave us and we threw away**: they close for good and cost nothing. **Two were questions we never asked**: the server does not offer the answer, so it has to be requested, and that costs a round trip which belongs on the failure path. That split is worth more as a method than as a taxonomy.

## What was wrong, and what it does now

**A directory that could not be listed arrived as an empty one.** FTP answers a LIST against a missing directory with a successful, empty listing, so "missing" and "empty" arrived identical.

The claim here is narrower than it first looked, and the narrowing came from measuring. A path that cannot be listed now **fails** rather than arriving as an empty directory, which is the property that matters: empty and absent stop coinciding, and a delete pass can no longer read a mistyped path as an emptied one. But on vsftpd the failure is generic, because that server answers a CWD into a missing directory with `550 Failed to change directory.`, the same words it uses for one you may not enter. The category is not on the wire, so it is not asserted. Not a display problem: a sync with `--delete` reads an empty source as "everything was removed" and plans to mirror that, so a mistyped path could authorise deleting a local tree. The provider now confirms existence when a listing comes back empty on an explicit non-root path. It costs nothing normally because it runs only in the ambiguous case, and it uses CWD, which needs no FEAT support and so behaves the same on servers advertising MLST, MLSD alone, or neither. The logic is not new: it is the CLI's own guard, moved to where all three surfaces reach it.

**The CLI planned deletions without the shared guard.** It already refused when the scan reported errors; the missing half was an empty source against a non-empty destination, which is exactly what the listing defect produced. A successful empty listing counts no errors, so it passed a check that existed. The core and the DAG both consulted the shared guard; this path restated half of it. The policy is now expressed over the facts it uses, so a caller holding maps neither materialises vectors nor copies the rule a fourth time. The frozen test that pins that policy is untouched, and a twin test compares the two shapes over the same eight cases on the whole `Result`, so the messages match byte for byte and not merely the Ok/Err.

**Downloading a file that does not exist hung.** Measured at 90 seconds to a client timeout and 600 through MCP, where it blocked the connection pool for everything else; rclone answers the same request in 1.3 seconds. The size lookup discarded its own failure with `unwrap_or(0)`, so a server that had just said "no such file" was answered with a fabricated zero and a RETR was issued anyway. That zero also fed the progress bar a false total and silently disabled intra-file parallelism, since zero is below every threshold. A reply naming a missing path is now an answer; every other size failure keeps the previous fallback, because servers refuse SIZE for legitimate reasons and refusing the download would break working setups. Both the download and the resume path are fixed: the resume is where the file is *more* likely to have gone, since time passes between the interrupted attempt and the retry.

**`stat` reported the listing's zero** instead of asking. It was right in the CLI, which hydrated it afterwards, and wrong in the GUI and MCP. The hydration is in the provider now.

**A 553 was retried three times.** Retry is decided by a classifier that had no rule for it, so it fell through to "unknown", which is retryable. It gets its own branch that records the one certain thing, that it is permanent, and does not claim a kind: RFC 959 spends 553 on "file name not allowed", and a 553 whose text does name a missing path is caught by the earlier branch and keeps the more precise kind. Only 553 is named, not "every 5xx": that classifier serves every provider, and in HTTP a 5xx is the retryable case, so a general rule would remove retry from every HTTP backend under load. A test pins that boundary.

**`mkdir` could not say "it is already there".** Every failure became one generic error, so the idempotent branch `mkdir -p` already had was unreachable on FTP: the variant it matches on was never produced, while five other providers produce it. rclone is idempotent here, so the target behaviour is not invented.

**A listing that changed directory restored it with the result thrown away.** A discarded restore is the worst available shape: the listing succeeds, the session is left elsewhere, and the damage surfaces later in an unrelated operation with nothing to connect it back. The restore now reports, the saved directory comes from PWD rather than from our own mirror of it, and the mirror follows the server immediately so it stays true even when the restore fails.

## The one that was not what it looked like

`mkdir` was written first to read the reply text, on the premise that "the server distinguishes these and we were discarding it". **That premise is false**, and a live server said so: vsftpd answers a duplicate `mkdir` with `550 Create directory operation failed.` and nothing else, the same text it uses for a refusal. The first version therefore fell through to "permission denied", replacing a vague answer with a wrong one.

So the directory is asked about instead. The question is put with CWD rather than with a stat, and that choice carries the cost of the branch: a stat is cheap only where MLST is advertised, and elsewhere falls back to listing the *parent* and searching it, which would make every duplicate `mkdir` cost a listing proportional to the parent and a recursive upload pay one per level of an existing tree. CWD is two commands regardless. It also answers a better question, because a CWD cannot succeed on a file, so "it exists and it is a directory" is the only thing success can mean: the guarantee stops being derived from a field and becomes intrinsic.

There is no text shortcut for servers that do word it clearly. One existed and was removed: it skipped the check, so a server saying `550 File exists` because a *file* holds the name would have been answered "the directory is already there", and an idempotent caller would have carried on to fail later somewhere else. Sparing those servers one round trip was not worth a guarantee that held on one branch and not the other, in a change that exists because a fix resting on the wording of a reply turned out to rest on nothing.

**Declared narrowing**: if the probe fails for a transient reason, a duplicate `mkdir` on a directory that really is there now returns the generic error rather than "already exists". Preferring not to assert is deliberate.

## Five compensations removed, four defects fixed

Not five and five: two of the compensations were the same defect. The second one says `mirrors ls` in its own comment, which is better evidence than anything that could be written here that the fix was in the wrong place, because a compensation that replicates itself from one chokepoint to the next has to be copied again for every new caller.

Each removal was checked in both directions: the compensation is gone from the CLI **and** its defect is fixed in the provider. Removing a guard without moving the fix would align the surfaces downwards, making the one that worked worse.

## Tests

Six unit tests and five integration tests, and the integration tests were **run**, against the repo's vsftpd fixture, not merely compiled: a missing directory answering as missing while a genuinely empty one still lists as empty; a listing leaving the working directory where it found it; `stat` reporting a real size; a duplicate `mkdir` being recognised; and a file occupying the name *not* being mistaken for an existing directory.

Running them is what found that the first `mkdir` fix did not work. No amount of reading would have: a correction that rests on the text of a server's reply cannot be validated by review, because the thing it depends on is not in the diff.

One unit test earned its place the hard way. Two earlier tests for the 553 rule passed with the rule deleted, because their fixtures contained words that matched an earlier branch: a rule with nothing holding it alive. The replacement uses a 553 carrying none of the other vocabulary, and it was checked by removing the rule and watching it go red.

**Not measured**: which text a given server sends for a bare 553. The exact wording is unverified; what is asserted is only that a reply carrying that status is permanent, which holds under every reading.

## A plaster, and it is labelled as one

The fixture container exits on its own mid-run (SIGSEGV, container status 139), and with no restart policy every test after that point failed with "Connection refused". Five red tests read as a regression in the code under test, which is the most expensive way to present an absent server, so the compose file now asks for `restart: on-failure` and the test file says what to check first.

That treats the symptom and hides the signal: a container that dies and comes back silently is a fault made quiet rather than made legible, which is the thing this whole change objects to. **The cause is open, not fixed.** The one clue available is that the stock vsftpd image has been up for 39 hours with no restarts while the image built from this repo dies within minutes, which says where to look and is not itself a diagnosis: nobody has investigated it. It is recorded as an open item so the restart policy is not mistaken for a resolution.

## What the review of this branch found

Three findings, all real, none applied in the form proposed.

**Both failures are kept.** When the hidden-file listing and the working-directory restore both failed, the listing error propagated first and the restore failure was never looked at, and the two fail together whenever the data connection drops. The finding asked for the restore to be checked first; reporting both is preferred, because choosing a winner discards the information the finding exists to preserve.

**A 550 on CWD no longer means "not found" by default**, and the server's own text is kept rather than replaced. A 550 is also what a server sends for a directory you may not enter, so reading every one as absent turned an inaccessible directory into a nonexistent one: the same mistake this change had already made on `mkdir` and had corrected by a live server, in a function forty lines away.

The fallback is `InvalidPath`, not `ServerError`, and that was found by writing the consequence down rather than citing it. `ServerError` maps to exit 10, which the retry predicate treats as retryable, so a directory that can never be entered would have been attempted three times: the very defect this change removes from the 553 path, reintroduced two functions away and noticed by nobody, including the review. `InvalidPath` is permanent on both retry paths, claims only that the path did not work, and is already what the MLST preflight in the same function returns for the same status.

**A byte count is not a reply code.** `mentions_ftp_status` accepted the digits after any colon, so "upload stalled: 553 bytes" was read as a permanent 553. The prefix must now be one of the `ProviderError` labels, a closed set defined in this repo, and an unrecognised prefix fails closed, leaving the error retryable. The boundary test that should have caught this checked only the form without a colon: a boundary test that looks at one side of the boundary. It now has four forms on each side.

## One visible change

Listing a path that does not exist on FTP exits **5** instead of 2, because the error is no longer classified as not-found. Anything checking exit 2 for "it is not there" on an FTP path sees a different code. Preserving the old one would mean asserting a category the server did not give us.

## Gates

`cargo fmt --manifest-path src-tauri/Cargo.toml --all -- --check`: 0.
`cargo clippy --manifest-path src-tauri/Cargo.toml --all-features --tests -- -D warnings`: 0.
`cargo clippy --manifest-path src-tauri/Cargo.toml --bin aeroftp-cli --all-features -- -D warnings`: 0.
Integration tests against the live fixture: 5 passed, 0 failed.
`cargo test --lib`: 3603 passed, 0 failed, 20 ignored. That run was started before the last two edits (the probe moving from stat to CWD, and the probe's restore failure no longer being collapsed), so it is reported for what it is rather than as a gate on this exact tree; it is running again and the figure is updated below when it lands. Neither edit is covered by a unit test, since both are on paths that need a server, and both are covered by the five integration tests above, which were run after them.
The full suite is CI's, on three platforms.

Tracker: #628

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved FTP/FTPS handling for missing paths, existing directories, failed uploads, and accurate file sizes.
  - Improved directory listings, including empty folders and working-directory restoration.
  - Prevented incomplete or empty synchronization scans from triggering deletions.
  - Improved error classification for FTP responses, permissions, rate limits, and misleading path text.
  - Clarified error messages with server responses and affected paths.

- **Tests**
  - Added coverage for FTP/FTPS path handling, directory operations, file sizes, synchronization safeguards, and server recovery scenarios.

### DIFF
--- a/src-tauri/src/bin/aeroftp_cli.rs
+++ b/src-tauri/src/bin/aeroftp_cli.rs
@@ -6469,12 +6469,7 @@ fn provider_error_to_exit_code(err: &ProviderError) -> i32 {
 }
 
 fn provider_error_message_looks_not_found(message: &str) -> bool {
-    let msg = message.to_ascii_lowercase();
-    msg.contains("no such file")
-        || msg.contains("not found")
-        || msg.contains("does not exist")
-        || msg.contains("file not exists")
-        || msg.contains("404")
+    ftp_client_gui_lib::providers::types::message_names_a_missing_path(message)
 }
 
 /// Reject a target file/folder name containing a character the active
@@ -8067,25 +8062,6 @@ fn print_profile_banner_once(name: &str, details: String, quiet: bool) {
     }
     if !PROFILE_INFO_PRINTED.swap(true, Ordering::Relaxed) {
         eprintln!("Using profile: {} ({})", name, details);
-    }
-}
-
-async fn maybe_hydrate_ftp_stat_size(
-    provider: &mut dyn StorageProvider,
-    path: &str,
-    entry: &mut RemoteEntry,
-) {
-    if entry.is_dir || entry.size > 0 {
-        return;
-    }
-    if !matches!(
-        provider.provider_type(),
-        ProviderType::Ftp | ProviderType::Ftps
-    ) {
-        return;
-    }
-    if let Ok(size) = provider.size(path).await {
-        entry.size = size;
     }
 }
 
@@ -15060,8 +15036,7 @@ async fn stat_tui_session_via_cli_handler(
     let provider = session.provider_mut();
 
     match provider.stat(&resolved_path).await {
-        Ok(mut entry) => {
-            maybe_hydrate_ftp_stat_size(provider, &resolved_path, &mut entry).await;
+        Ok(entry) => {
             let result_path = entry.path.clone();
             let result = cli_tui::worker::TuiStatResult {
                 name: sanitize_filename(&entry.name),
@@ -30114,28 +30089,6 @@ async fn collect_cli_list_entries(
         Err(err) => return Err(CliListError::Provider(err)),
     };
 
-    // FTP/FTPS disambiguation: some servers reply to LIST/MLSD on a missing
-    // directory with an empty listing instead of a 550 error, which collapses
-    // a missing path into an indistinguishable "empty directory" (exit 0).
-    // When the listing is empty and the user supplied an explicit non-root
-    // path, run a follow-up stat to confirm. If the path does not exist,
-    // surface NotFound with the correct exit code.
-    if entries.is_empty()
-        && !options.path.is_empty()
-        && options.path != "/"
-        && options.path != "."
-        && matches!(
-            provider.provider_type(),
-            ProviderType::Ftp | ProviderType::Ftps
-        )
-    {
-        if let Err(ProviderError::NotFound(_)) = provider.stat(&effective_path).await {
-            return Err(CliListError::Provider(ProviderError::NotFound(
-                options.path.to_string(),
-            )));
-        }
-    }
-
     // Filter hidden files
     let mut entries: Vec<RemoteEntry> = if options.all {
         entries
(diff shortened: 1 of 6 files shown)


axpnet:
@coderabbitai review

axpnet:
@coderabbitai review

axpnet:
@coderabbitai full review

axpnet:
@coderabbitai review

axpnet:
@coderabbitai full review

Issue #609, opened 2026-08-24
[BUG] Find Duplicates loading screen flickers in how many files it scanned


EhudKirsh:
**Describe the bug**
When Find Duplicates loads up to find non-identical duplicates, instead of the numerator and denominator steadily growing, they shrink a lot and then increase a lot every second or so.

**Steps To Reproduce**
1. Try to find non-identical duplicates in a folder that has at least hundreds of images.
2. The loading screen changes its numerator and denominator as described.

**Expected behavior**
The numerator and denominator should steadily grow and not ever decrease in value, never mind every second or so back and forth.

**Screenshots**

<img width="688" height="464" alt="Image" src="https://github.com/user-attachments/assets/1874641b-d443-44ae-a389-f0ae126576f2" />

<img width="692" height="472" alt="Image" src="https://github.com/user-attachments/assets/597b30cd-aba4-43c7-a672-3dda13f0ff98" />

**Environment:**
 - OS: Windows 10
 - AeroFTP Version: v4.1.7

**Additional context**
- I don't know if this is necessarily to do with non-identical duplicates. It might just be with identical duplicates as well if there are many like that in the same folder, but considering that identical duplicates are found instantly, it might be only to do with non-identical duplicates.

-  Please check AeroSync as well, because it also has this new loading page, and I think I remember that it might be affected as well.
	Any part of AeroFTP that has this new loading page.

axpnet:
Fixed on `main` for v4.1.8: each Duplicate Finder and AeroSync run now has its own operation ID, so stale overlapping events, results, and cleanup cannot replace the current scan. [Details and verification](https://github.com/axpdev-lab/aeroftp/issues/575#issuecomment-5402719725).

EhudKirsh:
Thanks, I tested Find Duplicates and AeroSync in v4.1.8 and can confirm that this is solved for both.

Just before closing this, I would like to suggest that you rename the 'Find Duplicates' window to 'Duplicates Finder'.
I think this is a good place to mention it, because you pretty much wrote it yourself. It sounds more natural to call it 'Finder' than 'Find'.
Please keep 'Duplicates' in the name instead of 'Duplicate' because the entire point is that there are multiple such files.
'Duplicates Finder' can remain as an option in AeroFile, it's just that its window I think should be renamed to 'Duplicates Finder'.
It gives flexibility of being able to call this feature by either name.

axpnet:
[...]

Issue #600, opened 2026-08-22
[BUG] Rclone Crypt in AeroFTP sometimes isn't decrypted by Rclone


EhudKirsh:
**Describe the bug**
I think that trying to encrypt with Rclone Crypt using AeroFTP currently doesn't always get decrypted in Rclone.

**Steps To Reproduce**
1. Set up Rclone Crypt on a cloud drive in AeroFTP.
	I set it on Yandex Disk OAuth 2.0 and used `/` as both the Remote Path and the Overlays Remote Path.
	I'm not sure if these change anything here or not. I didn't add any salt, and kept 'Standard (EME)' and 'Directory name encryption' ticked.
 
2. Encrypt a file or folder in that remote using AeroFTP.

3. Set up a Crypt remote in Rclone with that path and encryption password.
 
4. Running `rclone lsf YandexDiskCrypt:` resulted in this error message being printed for each item in that root directory:
	*Date* *Time* NOTICE: *ItemName*: Skipping undecryptable dir name: bad PKCS#7 padding - too long
 
5. Going to any GUI of Rclone simply shows that folder empty.

6. I created a new folder now in Jottacloud in AeroFTP and AeroFTP encrypted it in Rclone Crypt on the same `/` Remote Path and the Overlays Remote Path, and it was decrypted in Rclone. So I’m not sure what’s causing this, but I thought to report it even if I can’t fully repro a cause.

**Expected behavior**
To be able to set up Rclone Crypt in AeroFTP, encrypt items, and be able to decrypt them in Rclone consistently every time.

**Environment:**
 - OS: Windows 10
 - AeroFTP Version: v4.1.7

**Additional context**
- I can still see the items encrypted in YandexDisk: in Rclone, and the ‘🔒 RCLONE CRYPT’ button also works by toggling the showing of the plaintext and ciphertext names of items.
- As I wrote this, GitHub suggested that this might be a duplicate of #179.

axpnet:
The export correction was already on `main`; this pass adds a golden rclone v1.74.3 compatibility pin for an omitted `password2`, using the exact encrypted names produced by rclone. [Details and verification](https://github.com/axpdev-lab/aeroftp/issues/575#issuecomment-5402719725).

EhudKirsh:
[...]

Issue #453, opened 2026-07-23
[BUG] Manually sorting profiles in the GUI lands them in the wrong index number


EhudKirsh:
**Describe the bug**
Dragging and dropping profiles in the My Servers page still lands them in the wrong index number on several occasions. 

**Steps To Reproduce**
1. <ins>Grid view:</ins> dragging a profile to its left, up or to the very start or end works fine. But dragging it to the right or down results in an index number one smaller/higher than what it should be. When dragging a profile and dropping it on another, the dragged and dropped profile should inherit the exact index number as the profile it was dropped on, just like with `aero profiles --tui`> `#`.
 
2. <ins>Table view:</ins> sometimes holding on the ⋮⋮ of a profile doesn’t even manage to drag it.
	When it does, it always drops the profile 1 index smaller/higher than was indicated by the horizontal blue line.

**Expected behavior**
Dragged and dropped profiles should inherit the exact index number of where they’re dropped in the GUI, just like how
`aero profiles --tui`> `#` works well.

**Environment:**
 - OS: Windows 10
 - AeroFTP Version: v4.1.5, but it’s been this way for months, maybe the entire time.

**Additional context**
If it’s easier to script, or if you think it might look better or be more comfortable, you can adopt the way GitHub drags and drops repos in the profile page of users. Feel free to try it by dragging and dropping your repos in your GitHub profile https://github.com/axpnet to get a feel for it. You can see the ⋮⋮ at the top-right of each repo, which you can use to drag them. The nice thing is, since the GitHub website is in HTML, the same as the frontend of AeroFTP, it should be easy to check what it does using DevTools to learn from it. This isn't like trying to translate the behaviour of GTK or Qt or whatever else to HTML. You can do it even better than GitHub by not making the dragged rectangle so transparent and hard to read like GitHub does, and keeping Esc for cancelling the drag and drop instead of confirming it.

axpnet:
Thanks @EhudKirsh. Fixed on `main` for **v4.1.6**: drag-drop now lands on the drop target index.

Details and commit: tracker [#422](https://github.com/axpdev-lab/aeroftp/issues/422) queued row 62 (`5158375d`).

EhudKirsh:
Thanks, I see an improvement. However, two issues still remain, both in table view:
- Trying to move profile `#3` to be profile `#2` lands in `#1`. Basically, it still goes 1 higher or 1 index smaller than it should when going up.
[...]

PR #652, opened 2026-08-29
fix: close the v4.1.8 retests that did not hold

axpnet:
Closes the v4.1.8 items Ehud retested that did not hold. Rebased onto `f71fc5dd`.

## What this PR changes

- **#609** Find Duplicates window retitled Duplicates Finder (i18n; ru/uk already matched; sl left as-is — we do not read Slovenian well enough to replace one word on a dictionary argument).
- **#347** Permissions tab hydrates after `get_file_properties`; Filen versioning from REST then CLI fallback.
- **#397** Jottacloud trash: see the live measurements below. Help drops trash on pCloud Drive. CLI `trash` is still Nextcloud WebDAV only.
- **#600** rclone-crypt: reveal obscured password; empty `password2` is the default salt. Live vs rclone 1.74.3 both directions. Finding 4 declined — see explicit acceptance below.
- **#514** frontend overlay of live verify/compression/canary on `.aerosync`; `SyncTemplateProfile` now schemas those knobs so Rust import does not drop them.
- **#453** table DnD: selection cleared on dragstart, stable `data-server-id` on cells, sentinels always mounted. **Declared limit:** inspector cannot HTML5-drag. Verified: the table is the live view, 113 ids, grips present, synthetic `dragstart` fires. Unverified: a real WebKitGTK reorder, which needs a hand on a mouse.
- **#214 wiring / credentials:** imported Jottacloud refresh is bound via `extra.profile_id`. `apply_profile_options` stamps the record `id` last and skips options-borne `profile_id`/`profileId` (CWE-639). `auth_state` reads `jottacloud_refresh_<id>`, not only the singleton. Import dual-writes the singleton. If a walkthrough still names “unresolved credential selection”, that is this finding, already closed.

OneDrive trash: measured, not coded, tracker only.

## #397 live, stated precisely

**File restore is closed.** Three measured attempts on `ehud-397-0420.txt` on the rebuilt GUI (worktree debug binary, vault `aeroftp-dev-test`, Jotta via Cloud not WebDAV):

1. `?mv=` on `/Jotta/Trash/name` → 404, then `?mvDir=` with the trailing slash also 404.
2. `?restore=true` on the original Archive tombstone → 500 (retried 3×). Same class as stock 4.1.8 against the Trash URL.
3. rclone `?cphash=true` on that tombstone with JSize/JMd5/JCreated/JModified → 200. Log: `Restoring from trash via cphash: …/Archive/ehud-397-0420.txt?cphash=true (size=28 md5=9a6924f78982f7e759f371f08fb5d57e)` then `Restored from trash`. Trash 5→4; the file is back on Archive at 28 B.

`?mv=` and `?mvDir=` both 404 because **the Trash path is not an object**, the listing being a virtual view. The trailing-slash fix the earlier review asked for was a correct observation about a mechanism that does not apply here. The reviewer was right that the code was wrong and wrong about what would fix it.

**Folders.** The classification now takes the folder branch, proven live by the log reading `?restore=true` rather than cphash, and JFS then answered **500** three times and the folder did not come back. So: correctly classified, and no working JFS command exists for it. Nested trash entries keep `<abspath>` (accumulated and entity-decoded, not assigned per XML text chunk) so restore uses the original parent, not the mountpoint root.

**Deleted column, both halves.** The width squeeze truncated `Aug 29, 2026, 15:45` to `Aug 29, 2026, 1` (what the owner saw). Separately the column showed revision mtime, so `Documenti.aerovault` read June 24 instead of July 27: the value was **wrong**, not merely clipped.

## Explicit acceptance: rclone-obscure finding 4 (F-652-4 sharpening)

rclone obscuring is AES-256-CTR with **no MAC**. Nothing in the value distinguishes `obscure("")` from a 22-character URL-safe Base64 string that is only an IV: they decode to the same shape. The collision is not only `A`×22. Any 22-character password in `[A-Za-z0-9_-]` (password-manager alphabets hit this) reveals as empty and the derived key opens nothing, silently. Longer lengths in that alphabet can yield short random ASCII as well. Refusing empty reveals brings #600 back. Treating every IV-only decode as empty turns those 22-character passwords into the empty key. Declining to change the KDF on a collision that cannot be resolved is the call.

A control next to the field saying the value is rclone-obscured is available on the paste path. That does not need the value to carry provenance. **Not this PR.** It is a UI change and an owner product decision. The remedy exists; the blocker is a product choice, not a cryptographic impossibility. Rejecting an empty reveal for the password (not the salt) is optional and still not this PR.

## Merge

Do not merge on this branch until the walkthrough’s remaining bounded risks are fixed or explicitly accepted as above. Folder restore has no working JFS command (declared). #453 real-mouse reorder is unverified (declared). Finding 4 is explicitly accepted.

## Summary by CodeRabbit

* **New Features**
  * Sync templates now preserve verification, compression, and canary settings.
  * Saved server profiles retain profile-specific authentication across reconnects and imports.
  * Storage quota displays include versioning usage where available.
  * Filen quota information uses the service API first, with fallback support.

* **Bug Fixes**
  * Prevented caller-supplied profile options from overriding saved profile identities.
  * Improved Jottacloud rename and trash restoration.
  * Improved rclone credential handling.
  * More reliable server drag-and-drop and file properties display.
  * pCloud no longer shows unsupported Trash options.
  * Improved Trash table readability and Jottacloud Trash panel layout.

* **Localization**
  * Updated the Duplicate Finder title across supported languages.

## Adversarial review of the delta, and what it found

The two commits that followed the last automated review were reviewed adversarially. One substantive defect came out of it, with an experiment attached rather than an opinion, and it is fixed here.

**`parse_trash_xml` lost the spaces around an XML entity.** The reader was configured with `trim_text(true)`, which trims every Text EVENT rather than the content of the node. An entity splits `<abspath>/user/Jotta/Archive/Photos &amp; Videos</abspath>` into three events whose own edges carry the spaces, so accumulating them produced `Photos&Videos`: `trash_relative_path` then built a parent that does not exist and a restore would GET the wrong path. The comment above that line described the intended remedy ("accumulated, then trimmed once") rather than the code, which is the worst shape a comment can take: precise about something that does not happen, so the reader believes it has already been checked. Trimming is now off, the three scalars that need it keep trimming themselves, and `parse_trash_xml_keeps_spaces_around_an_entity_in_abspath` pins it. That test was run against the defective code first and observed failing (`/Photos&Videos/x.txt` against `/Photos & Videos/x.txt`).

**A comment promised a surface that does not exist.** It said the Trash-only rename is "called by a dedicated surface". Nothing calls `rename_in_trash`: no Tauri command, no button in the Jottacloud trash manager. The comment now says so.

## What the tests here do NOT pin, stated rather than counted

`extras_for_connect_overwrites_options_borne_profile_id` and the tombstone test pin their helpers, not the sites that call them: if `connect_provider` went back to `or_insert`, or if the tombstone check stopped being consulted, both tests would stay green. The call site sits inside `connect_provider`, which resolves a credential from the vault and ends in `provider.connect().await`, so exercising it needs a real connection unless the config assembly is first extracted into a pure function. That refactor is possible and not hard; it is left out to keep this change from growing, and this paragraph is here so the cost is visible rather than implied.

## One deliberate exception to a house rule

`rename_in_trash` and its helper have no callers and CLAUDE.md says to remove dead code immediately. They are kept, with the reason and the expiry written at the call-site comment: #397 reports exactly that operation failing for a user, this is its implementation with its tests, and deleting it would throw away the only answer to that report. If #397 does not wire it during this release cycle, it goes.

This PR is one part of #397, not all of it: the Dropbox tier and OneDrive cases in that issue's title are untouched, and so is the request to say in Help > Providers that pCloud does not allow trash management over OAuth.


axpnet:
@coderabbitai full review

axpnet:
Head moved to `8ce4aa0414cc37dfd73e6db8ca9acd5b353e3dd8` (2026-08-30 01:55 CEST), so the review record on this pull request is no longer on the head and has to be reconfirmed or redone on the new sha before any merge. I am writing that here rather than relying on anyone remembering it: a real review on a sha that is no longer the head looks exactly like a review on the head, and that mistake is currently holding another pull request of ours.

What the new commit contains: the trim fix in `parse_trash_xml`, its test (run against the defective code first and observed failing), and two comment corrections. Gates on this sha before pushing: `cargo fmt --all -- --check` clean, `cargo clippy --all-targets -- -D warnings` clean, and the full `cargo test --lib`, 3637 passed, 0 failed, 20 ignored.

The body now states what these tests do NOT pin, rather than leaving the count to imply coverage.

axpnet:
Review record on the current head, posted here because it was given to me in a side channel and a record that lives outside the pull request is invisible to whoever opens the pull request to merge it. I am the author of this change, so to be explicit: this verdict is not mine. It comes from the adversarial review session that found the trim defect in the first place, and I am transcribing it, not endorsing my own work.

Verdict on head `8ce4aa0414cc37dfd73e6db8ca9acd5b353e3dd8`, delta `40a334160..8ce4aa041` read in full: nothing to report.

Detail as given. The trim fix is closed: `trim_text(false)` at `jottacloud.rs:2305`; the scalars (`size`, `state`, `modified`/`updated`) trim themselves at 2525 and are guarded by `current_tag`, which is cleared on `End` at 2510, so the whitespace-only Text events that now arrive fall into the `_` arm or outside both branches; the `abspath` arm at 2513-2523 accumulates everything and the single trim stays at the point of use. `parse_trash_xml_keeps_spaces_around_an_entity_in_abspath` at 2813-2832 counts as a pin because it was observed failing on the defective code. The two reader configurations at 823 and 2019 stay `trim_text(true)` and are not defects there.

On the unwired `rename_in_trash`: the comment at 1580-1593 now says the truth, gives the reason and gives the expiry. One reservation, explicitly non-blocking: an expiry written in a comment does not enforce itself, so it should also exist as a line in the release tracker, otherwise nobody rereads it a cycle from now. I agree with that and it is recorded for the resume rather than done now, because the team is in a coordinated stop.

Earlier findings from the same review on the previous head, for the record: no second door for `profile_id` beyond the one declared, the tombstone signal is the right one with a stated reservation about homonymous tombstones, and two of the tests pin their helper rather than the call site, which the body now states instead of leaving the test count to imply coverage.

PR #653, opened 2026-08-29
fix(ftp): read what the listing actually says

axpnet:
Five defects in how FTP listings are read, all of them cases where the parser knew less than it acted on.

**The parser was chosen by failure, not by shape.** Dispatch tried one dialect, and fell through to the next when it returned nothing. A line that parses as neither dialect and a line that parses as an empty result are the same event to that logic, so a server whose format we do not handle silently became a server with nothing in it. Dispatch now decides on the shape of the line.

**Names lost their spaces.** A filename with spaces was truncated at the first one, because the field split assumed a fixed token count. The name the server sent is kept as the server sent it.

**Unreadable rows are now reported.** They were dropped by a `filter_map`, so a listing where some rows failed to parse looked exactly like a listing where those files do not exist. The count and the first offending row are carried out.

**One answer for a size that cannot be read**, rather than each caller inventing its own.

**And a sixth item, found by the rebase and declared rather than folded in.** A listing can now come back empty for two different reasons: the directory is empty, or every row was unreadable. In the second case the existence check passes, the directory does exist, and we return an empty list, which is to say "this directory is empty" when the truth is "we could not read a single row".

No data was ever at risk from it: `orphan_delete_guard` already refuses a delete planned from a source that came back empty, in the direction where that matters. What was wrong is what the refusal *says*, since it reports the source as empty and sends the reader to the wrong conclusion. And `ScanCompleteness` counted zero listing errors for such a listing, because it increments only on `Err` and a listing with every row dropped is an `Ok`, so a scan that read nothing was recorded as complete.

The partial case is left open and stated rather than half-solved: if one row in ten parses, the list is not empty, no guard fires, and the nine lost rows read downstream exactly like deletions. Closing it means the provider reporting completeness alongside the entries, which is a trait implemented by 31 providers.

**Verification.** `cargo fmt --check` clean, `clippy --all-features --tests` and `clippy --bin aeroftp-cli` both zero, `cargo test --lib` 3622 passed with 0 failed. The two frozen characterisation baselines pass **without regeneration** after the rebase onto `f71fc5dd`, which is the free test on #651: it changed provider behaviour and did not touch the parser, as it declared.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved FTP directory listing compatibility across Unix, DOS, and MLSD formats.
  * Preserved consecutive, embedded, and trailing spaces in filenames.
  * Correctly handles DOS listings with multiword filenames and unknown file sizes.
  * Reports unreadable listing rows instead of silently omitting them or showing an empty directory.
  * Improved warnings and error handling when FTP listings cannot be fully read.
  * Better recognizes directory listing headers containing absolute paths and spaces.


axpnet:
@coderabbitai review

PR #654, opened 2026-08-29
refactor(providers): align every recursive descent on the walkable-dir contract

axpnet:
**This changes no behaviour, on any of the ten sites.** Every one of them already produced the same answer it produces now. What changes is that the contract at `RemoteEntry::is_walkable_dir` (`providers/types.rs`) becomes checkable by machine, instead of being a rule with ten documented exceptions that nobody can verify without reading all ten.

**The check, and the numbers I actually ran.** `git grep -n -A6 -E 'if [A-Za-z_][A-Za-z0-9_]*\.is_dir\b' -- src-tauri/src/providers | grep -E 'Box::pin\(|\.push\('` matches thirteen lines, across ten distinct sites, on this PR's base, and nothing on this head. The scope of that claim is exactly `providers/`: the same pattern lives on in `bin/`, `ai_core/`, `sync_core/` and `lib.rs`, where a wider scan reports 35 further candidates, three of which I read and all three were real descents. This PR does not touch them, and the sentence above is not a statement about the repository.

**The first version of this PR made that claim while four sites still failed it**, which is the same error in the opposite direction to the one corrected further down: there I verified a premise on a path I was not touching, here I made a claim about a whole directory and verified it only where I had touched. Both are a gap between the scope of the assertion and the scope of the check.

**The four, and why each is cosmetic today.** `internxt` writes `is_symlink: false` at all five of its construction sites and `filen` at both of its, so the flag cannot ever be true there. The GitHub entries that reach `repo_mode` come from a constructor where `is_dir` is `content_type == "dir"` and `is_symlink` is `content_type == "symlink"`, two comparisons of one field over disjoint values, so the two are never true together. The fourth is the instructive one: SFTP `rmdir_recursive` already handles symlinks in an explicit arm before the directory arm, so it was equivalent to the contract and semantically had nothing wrong with it, but a grep cannot know that, and one site that needs a human reading costs the property for all of them.

The first version of this change claimed a defect in the SFTP `find` walk and was wrong. The correction is recorded here rather than quietly dropped, because the reason it looked like a defect is the useful part.

**Why `find` was never unsafe.** It reads its entries from `read_dir`, whose attributes carry the entry's own mode (lstat semantics; the `list` path states this where it spends an LSTAT only when the server omits the type bits). A symlink therefore arrives with `S_IFLNK`, `metadata_to_entry` derives `is_dir` from `S_IFDIR`, and the answer is false. The walk could not descend into a link, so the guard there is exactly as cosmetic as the others.

**Where the measurement came from, and why it did not apply.** A live SFTP fixture with `linkdir -> realdir` reports `is_dir=true is_symlink=true link_target=Some("realdir")` for the link. That is real and it is pinned by the ignored test in `tests/live_sftp_symlink_contract.rs`. But it was measured through `list()`, which is the one path that resolves the target and sets `is_symlink` separately. `find` does neither. The premise was measured carefully on the wrong path.

**Three facts about this tree, which is the part worth keeping.** `is_walkable_dir()` differs from `is_dir` on exactly one provider, SFTP, and only on the `list()` path, because that is the only place where the two flags are derived from two different facts. Everywhere else they come from one: FTP reads `d` and `l` from the same leading character of the same permission string, and on MLSD from the same `type` fact over disjoint values; GitHub compares one `content_type` field twice. WebDAV, Azure, internxt and filen write `is_symlink: false` at every construction site, so the guard there cannot protect anything, because a guard is worth exactly what the data feeding it is worth.

**The test characterises the assumption, not the behaviour.** `a_symlink_from_readdir_is_not_a_directory` passes today and proves nothing new. It exists for the day somebody makes `find` resolve the target the way `list` does: at that moment `is_dir` becomes true for a symlink to a directory, the guard stops being cosmetic and starts carrying weight, and the assertion goes red instead of leaving the change to be found by a user whose recursive search followed a link to an ancestor.

Out of scope, with the reason: `sync_core/scan.rs` has two further descents on a raw `is_dir`, and closing them REMOVES entries from the remote scan. On `--direction download --delete` the remote is the source, so entries that stop being enumerated become orphans and are deleted from the user's disk on the first run after the upgrade, while the same edit is safe on `upload`. That needs the delete direction in the reasoning, a test per direction, and an answer for trees already synced under today's behaviour.

### DIFF
--- a/src-tauri/src/providers/azure.rs
+++ b/src-tauri/src/providers/azure.rs
@@ -1570,7 +1570,7 @@ impl StorageProvider for AzureProvider {
         }
         let entries = self.list(path).await?;
         for entry in entries {
-            if entry.is_dir {
+            if entry.is_walkable_dir() {
                 Box::pin(self.rmdir_recursive(&entry.path)).await?;
             } else {
                 self.delete(&entry.path).await?;
--- a/src-tauri/src/providers/filen/mod.rs
+++ b/src-tauri/src/providers/filen/mod.rs
@@ -2845,7 +2845,7 @@ impl StorageProvider for FilenProvider {
                         if crate::providers::matches_find_pattern(&entry.name, pattern) {
                             results.push(entry.clone());
                         }
-                        if entry.is_dir && results.len() < 500 {
+                        if entry.is_walkable_dir() && results.len() < 500 {
                             next_dirs.push(entry.path.clone());
                         }
                     }
--- a/src-tauri/src/providers/ftp.rs
+++ b/src-tauri/src/providers/ftp.rs
@@ -1349,7 +1349,7 @@ impl StorageProvider for FtpProvider {
 
         // Delete contents first
         for entry in entries {
-            if entry.is_dir {
+            if entry.is_walkable_dir() {
                 // Use Box::pin for recursive async call
                 Box::pin(self.rmdir_recursive(&entry.path)).await?;
             } else {
@@ -1453,7 +1453,7 @@ impl StorageProvider for FtpProvider {
             self.current_path = saved;
 
             for entry in entries {
-                if entry.is_dir {
+                if entry.is_walkable_dir() {
                     dirs_to_scan.push(entry.path.clone());
                 }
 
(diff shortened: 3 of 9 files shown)


axpnet:
@coderabbitai review

axpnet:
@coderabbitai full review

axpnet:
@coderabbitai full review

PR #655, opened 2026-08-29
fix(ftp): stop waiting for bytes the server already refused to send

axpnet:
## What this fixes

A `get` of a file that is not there could wait for ever, and "for ever" is meant literally rather than as shorthand for slow.

Captured on the wire, twice, on a remote server, with a stable picture between the two snapshots:

```
ESTAB  Recv-Q=55  ->  <server>:21     (control)
ESTAB  Recv-Q=0   ->  <server>:30071  (data)
```

The control socket held 55 unread bytes with nothing received for 22 seconds. The data socket was open with 388 bytes sent and nothing ever coming back. Neither socket had a timer armed, and the timer column was verified to work by checking that 40 other sockets on the same machine did print one, so the emptiness is a fact rather than a limit of the tool. No kernel event could ever end that wait.

The server had already answered. Its `550` was sitting in the client's own receive buffer while the client waited on the data channel, because reading the data channel means not reading the control channel. The information was not on the wire and not in a struct: it was inside the client process, in the client's socket, and nothing was looking at it.

## Why the fixture cannot show it

On loopback the same code returns in two seconds. There the data socket closes at once, the wait ends by itself, and the ordinary path finds the `550` afterwards. Same server software, same version, opposite outcome, decided by which side closes first.

The docker fixture in this repo therefore cannot exercise this defect, and a green run against it proves nothing here. That is said next to the tests as well as in this description, because the tests are where somebody will read a green and draw a conclusion.

The two tests build the failing state instead: a data channel that is open and silent, plus a refusal already sitting unread in the control socket. Without the fix the first test times out; that was confirmed by removing the fix and watching it fail, not assumed.

## The part that was not in the plan

Noticing that the server spoke is not enough, and the obvious version of this change is dangerous.

The completion reply arrives on the same channel that is now being watched, so treating any reply as a refusal would abort a transfer that was working and truncate the file. The first digit is read with `peek`, which does not consume: `4` and `5` end the wait, anything else stops the watch and leaves the reply in place for `finalize_retr_stream`.

A remedy for a hang that silently truncated files would have been the worse defect of the two.

## Cancellation safety is load-bearing

Only the readiness is raced. `readable()` and `peek()` take `&self` and consume nothing, so losing the race costs nothing.

`read_response_in` is not cancellation-safe: it reads lines from a `BufReader`, and dropping it mid-line leaves the control channel misaligned. It is therefore called after the race, never inside it. Moving it into the `select!` would look like a simplification and would trade a control channel the client ignores for one it corrupts. That constraint is written above the function, not only here.

## An encrypted control channel

Reading the first byte of a reply only works in the clear, and that is the protocol rather than a limit found by measuring: a TLS record opens with a ContentType in 0x14..0x18, while ASCII `4` and `5` are 0x34 and 0x35, so the sets do not intersect and no server can put a reply's digits there. Under FTPS the classification above cannot fire, and the default TLS mode is the one most users get.

What survives encryption is the READINESS. The server spoke, even if what it said cannot be read from this side of the TLS layer, and during a healthy transfer it has no reason to: the preliminary reply was consumed when the data channel opened and the completion reply comes after the data are done.

So a reply arriving mid-transfer SHORTENS the deadline instead of ending the wait, from thirty minutes to thirty seconds, and when the shorter deadline expires the pending reply is read, so the user still gets the server's own words a few seconds later rather than a bare timeout. The difference between a clear and an encrypted control channel becomes the latency of the answer, not whether there is one.

It shortens rather than aborts on purpose, and the shortened deadline is inactivity rather than a grace period. A transfer that is ending well keeps delivering the bytes already in flight, each one resets the timer, and this never fires: it fires only when the server has spoken AND the data have stopped, which are never both true in a healthy transfer. Arming on readiness alone would cut short a transfer that was finishing.

## Leaving a transfer without leaving a mess

Every early exit from a data loop used to drop the `DataStream` without calling `finalize`, so inside the crate `data_connection_open` stayed true and the completion reply stayed unread. What that meant was decided by whatever came next: another data command is refused with "data connection is already open", which is recognised and reconnects, but a CONTROL command reads the stale `226 Transfer complete` as its own reply and fails while quoting the success of something else.

This is not a reading of a third-party crate. This product has had the defect and named it. `FtpManager` sends a NOOP before every LIST, and its comment says why: "drain any pending control-connection responses from a previous large transfer/listing. This prevents stale data from leaking into the next LIST response (phantom files bug)." That mitigation exists only on the twin. The path serving the GUI and the CLI has no equivalent.

Those exits were rare before this change and are not any more, because a deadline and a refusal were added to four loops. A defect made common by a change belongs to that change, so the eight exits now share one way out: `abort` under a five second budget, and if the server will not answer even that, the connection is dropped rather than reused. It reports which of the two happened, because `list_inner_opts` carries on to its LIST branch after MLSD fails, and a silently discarded session would turn a server that advertises MLSD without honouring it into `NotConnected`.

One reply needs naming, because reading it wrongly was about to cost a reconnect for every missing file. After a refusal has been read there is no transfer left to abort, and servers say so with 225: pyftpdlib answers "225 No transfer to abort." and vsftpd "225 No transfer to ABOR.", both measured. `abort` waits for 226 or 426, so it reports that as an error.

It is not one. `abort` sends ABOR, drops the stream, CLEARS the connection flag, and only then reads the reply, and the reader consumes the line before deciding whether to accept its code. On a 225 the flag is already down, the line is already gone, and the control channel is aligned: the session is clean. Treating every error from `abort` as a broken session would have discarded it, and on the pooled providers behind the GUI, the MCP server and the TUI that is a fresh TLS handshake per missing file, which is the commonest case this change touches.

That cost would have been introduced here rather than inherited, and it was nearly written into this description as a limitation of the crate with the remedy deferred. A limit recorded against someone else's library is one people stop looking at, with a clear conscience because the reason is written down, and this one was a single line in this change.

Curing it at the exit, rather than draining before the next question, is the difference from the NOOP. The NOOP is what has held this together until now.

`disconnect` is bounded for the same reason. It sends QUIT and then reads the reply with no deadline, and it is reached from seven places through `reconnect_after_data_error`, all of them after something has already gone wrong: on a server that had stopped answering, the remedy meant to throw away a poisoned connection hung on the goodbye instead. The `let _ =` in front of it discarded the result, not the wait. The twin has bounded this for a long time, with the same five seconds used here.

WHAT THE CLEANUP DOES NOT COVER, said plainly because a function that looks complete is read as complete: it runs on exits that RETURN. A future dropped from outside never reaches it, since dropping an async fn runs the destructors of its locals and none of the code that follows. The transfer executor wraps a download in `tokio::time::timeout` and lets the future fall on expiry, so on that path the session is left in exactly the state this exists to prevent. Moving that deadline inside the data loop would turn it into an ordinary error return and bring it back through here, which is where the shared primitive will put it anyway.

## What this does NOT close

This closes four of the five RETR loops this provider drives itself. It does not close the class, and three of the gaps are inside FTP rather than beyond it.

The fifth loop is `ftp_download_one_range`, the parallel range download. Its read sits in a `select!` with the cancellation token alone: no deadline, no control watching, no cleanup on the way out. It is the only one of the five that dials its own connection, created and dropped inside the function, so the stale-session problem below does not reach it; the unbounded wait does.

`STOR` is a loop this provider drives too, and it gets neither the deadline nor the watch here. The failure is the mirror image: the server stops reading, `write_all` blocks once the kernel buffer fills, and the reply sits unread. `APPE` is worse, because `resume_upload` still calls `append_file` with a bare `map_err`, so its message carries no path, its refusals get no permanence classification, and it has no deadline at all. That is the branch the transfer executor takes when it retries from a partial offset, which is exactly where a pointless retry costs most. Both are tracked, and both belong with the shared data-loop primitive rather than with a sixth loop rewritten inside this change.

The `LIST` data loop lives inside `suppaftp`: `list()` reaches `get_lines_from_stream`, which loops on `read_until` with no deadline and no control watching. The helper in this provider cannot reach it, because the loop belongs to the crate. `MLSD` reaches the same loop, and `MLSD` is the branch tried first.

That is not a guess about a related path. It is the same failure, measured on the same remote server, with the passive port opened by hand and the data channel inspected without blocking:

```
MLSD /nope-xyz    control: "501 No such directory."          instant   data: OPEN AND SILENT
RETR /nope.bin    control: "550 No such file or directory."  instant   data: OPEN AND SILENT
MLSD /            control: 125 then 226                                data: 250 bytes
```

The first two rows are one defect written twice. The reply is already on the control channel, the data channel is open and will never deliver anything, and nothing is watching. The second row is what this pull request closes. The first is not touched by it.

What prevents a user from hanging there today is one thing only: the MLST probe, which runs before MLSD and answers instantly on a missing path. Measured end to end against that server, `ls` on a directory that is not there returns in 0.48 seconds. But the probe is gated on `mlst_supported`, so a server that speaks MLSD and does not advertise MLST reaches the MLSD call, opens the data channel, and waits with neither of the two remedies in this change.

This was declared open in an earlier draft of this description as a matter of imprecise categorisation. That was an understatement, corrected here after measuring it: it is the same permanent hang, on the branch modern servers take first. It is pre-existing, this change does not make it worse, and it is tracked.

The anti-hang probe that covers listings is gated on `mlst_supported`, and vsftpd 3.0.5 answers `500 Unknown command` to `MLST` (measured). So on servers that do not advertise MLST the listing has no anti-hang defence at all. This is a pre-existing gap rather than something introduced here, and it was found by asking a server rather than by reading the code: the condition is written in plain sight, and what nobody had asked was how often it is false.

Neither fixture in this repository can exercise any of that, and this is measured rather than assumed. Both FTP containers are vsftpd, and `FEAT` on each answers with `MDTM` and `SIZE` only: no `MLSD`, no `MLST`. So the primary listing branch and the anti-hang probe that is supposed to cover it are both unreachable in the current lab, which is a large part of why this class survived. Closing it needs a ProFTPD or pure-ftpd fixture first, not code.

Whether a `LIST` actually hangs on such a server is NOT measured. I have no server that exercises it, and asserting it without having seen it would be the same overclaiming this branch has spent its time removing.

The remedy is plausibly small rather than structural, and three things about it are worth recording before anyone attempts it:

The reconnect machinery already exists and already wraps `list` and `list_for_delete`: `reconnect_after_data_error` disconnects and reconnects, so it discards a misaligned connection rather than reusing it. That was my main objection and it does not stand.

But `is_stale_data_connection_error` recognises staleness by SUBSTRING, not by type, so a timeout error saying "no data arrived" matches none of its needles and would fall through without ever reaching the reconnect. Making it match by adding a needle is the same text-matching class this branch has been removing in three other places.

And the reconnect retries once, so the real ceiling is twice the timeout, not once. For a listing a TOTAL cap is the right shape, the opposite of the choice made below for transfers, because the correct form follows from whether the operation has an intrinsic size bound: a listing does, a download does not.

## Retries

A refused data command is no longer retried when the refusal is permanent.

Measured against vsftpd 3.0.5: `put` into a directory that does not exist is refused with `553 Could not create file.` and `get` of a missing file with `550 Failed to open file.`, both instantly, and both were then attempted twice more for nothing.

The status decides permanence, not the text: RFC 959 makes 5yz permanent and 4yz transient, the same rule already applied to `CWD` on the branch below this one.

The obvious fix would have been wrong. Neither reply names a missing path: they say the operation did not happen, and a server sends them both for a path that is absent and for one it will not let you touch. Adding them to the missing-path vocabulary would state something the server did not, which is precisely the defect removed from the `550` on `CWD`.

## Messages

`put` says again what to do about a missing parent. The CLI used to check the parent before every upload and name the `mkdir -p` that would fix it; that check was removed when the provider began classifying these failures, and for the other four removed checks the reasoning held. Here it did not, because a classifier reads a reply while what was lost was a sentence the user could act on.

The parent is now asked about AFTER a permanent refusal rather than before every upload, so the successful path pays nothing, and it lives in the provider, so the GUI and the MCP server get the sentence the CLI-only check never gave them. If the parent turns out to exist, the original error is returned untouched.

All five `RETR` sites also name the path now. They were five identical copies of the same `map_err`, and none of them said which file had failed.

## One pinned expectation changed

`a_store_into_a_missing_directory_is_not_found_not_a_transfer_failure` said "must not become NotFound" in its message and asserted the `TransferFailed` variant, which is a narrower claim than the one it named. `InvalidPath` satisfies the property; the row now asserts both, and the reason the variant moved is recorded beside it.

It was sharpened rather than relaxed. Relaxing an assertion until it passes destroys the only record of what the test was for.

## One commit rather than three

The three parts are separable in principle, but they live in one file and reconstructing the intermediate trees by hand risks committing one that does not build, which costs more than the attribution gains. Each part is named in the commit message and commented where it lives.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved FTP transfer error messages with clearer operation and file path details.
  - More reliably detects refused transfers, timeouts, missing directories, and unreadable listings.
  - Improved downloads, uploads, resumed transfers, range requests, and hidden-file deletion.
  - Added safeguards to prevent transfers from hanging when servers stop responding.
  - FTP listings now explicitly report empty or inaccessible directory results.

### DIFF
--- a/src-tauri/src/providers/ftp.rs
+++ b/src-tauri/src/providers/ftp.rs
@@ -1360,12 +1360,7 @@ impl StorageProvider for FtpProvider {
             let n = match step {
                 DataStep::Read(n) => n,
                 DataStep::ControlRefused => {
-                    let stream = self.stream.as_mut().ok_or(ProviderError::NotConnected)?;
-                    let refusal = stream
-                        .read_response_in(&[Status::ClosingDataConnection])
-                        .await
-                        .err()
-                        .unwrap_or(FtpError::BadResponse);
+                    let refusal = self.read_queued_refusal().await?;
                     let _ = self.abandon_transfer(data_stream).await;
                     return Err(Self::classify_data_failure("reading", remote_path, refusal));
                 }
@@ -1740,12 +1735,7 @@ impl StorageProvider for FtpProvider {
             let n = match step {
                 DataStep::Read(n) => n,
                 DataStep::ControlRefused => {
-                    let stream = self.stream.as_mut().ok_or(ProviderError::NotConnected)?;
-                    let refusal = stream
-                        .read_response_in(&[Status::ClosingDataConnection])
-                        .await
-                        .err()
-                        .unwrap_or(FtpError::BadResponse);
+                    let refusal = self.read_queued_refusal().await?;
                     let _ = self.abandon_transfer(data_stream).await;
                     return Err(Self::classify_data_failure(
                         "resuming",
@@ -2054,12 +2044,7 @@ impl StorageProvider for FtpProvider {
             let n = match step {
                 DataStep::Read(n) => n,
                 DataStep::ControlRefused => {
-                    let stream = self.stream.as_mut().ok_or(ProviderError::NotConnected)?;
-                    let refusal = stream
-                        .read_response_in(&[Status::ClosingDataConnection])
-                        .await
-                        .err()
-                        .unwrap_or(FtpError::BadResponse);
+                    let refusal = self.read_queued_refusal().await?;
                     let _ = self.abandon_transfer(data_stream).await;
                     return Err(Self::classify_data_failure(
                         "reading a range of",
@@ -2201,6 +2186,7 @@ enum ControlWatch {
 }
 
 /// What one turn of a data-reading loop produced.
+#[derive(Debug)]
 enum DataStep {
     /// Bytes from the data channel; `0` is end of stream.
     Read(usize),
@@ -2276,6 +2262,30 @@ impl FtpProvider {
             self.stream = None;
         }
         verdict
+    }
+
+    /// Read the refusal the server has queued, without waiting for one that
+    /// may not be there.
+    ///
+    /// Reached only after a peek has seen bytes on the control channel, so a
+    /// reply is expected. Expected is not guaranteed: a partial line, or a TLS
+    /// record carrying something other than a complete reply, leaves
+    /// `read_response_in` waiting, and it has no deadline of its own. Every
+    /// wait on this path is bounded now, including the one that reports why the
+    /// other ones ended.
+    ///
+    /// One function rather than four copies at the sites that need it.
+    async fn read_queued_refusal(&mut self) -> Result<FtpError, ProviderError> {
+        const REFUSAL_BUDGET: std::time::Duration = std::time::Duration::from_secs(5);
+        let stream = self.stream.as_mut().ok_or(ProviderError::NotConnected)?;
+        let read = stream.read_response_in(&[Status::ClosingDataConnection]);
+        Ok(match tokio::time::timeout(REFUSAL_BUDGET, read).await {
+            // A refusal, which is what this path exists to report.
+            Ok(Err(err)) => err,
+            // A positive reply, or nothing at all in time: neither is a
+            // refusal, and neither is worth waiting past its budget for.
+            _ => FtpError::BadResponse,
+        })
     }
 
     /// Decide whether a session survives, from what `abort` said about it.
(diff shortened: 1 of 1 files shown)


axpnet:
@coderabbitai review

Issue #658, opened 2026-08-30
[BUG] AeroFTP crashes without error, while trying to start uploading large files


eagle-te:
searched for a similar existing tickets or known issues, but didn't found one

**Describe the bug**
While trying to upload a large file (25.2GB) the application crashes without error and no log file.
No log line with error/warn was shown in the activiy log or debug logs windows.
No log file at `%appdata%/aeroftp`.

**Steps To Reproduce**
1. Open a sftp connection to a Hetzner storage box
2. Start a transfer a file with a size of at least 25.2GB(smaller ones at 3.8GB, 2.4GB worked fine), Transfermode: Compatibility 1x
3. See the application trying to start the transfer
4. Application hard crashes

**Expected behavior**
Uploading the file as expected.
I now tested the upload of the same file with winscp, which worked.

**Environment:**
 - OS: Windows 10
 - AeroFTP Version: 4.1.8 (installed via winget)
 - linked libraries:
   - russh: 0.62.5
   - russh-sftp: 2.4.0

**Tested also on linux**
- OS: Manjaro, kernel 7.2.0
- AeroFTP: 4.1.8

Did the same test, with a slightly larger file. What I observed is, that aeroftp takes very long trying to start the upload. htop showed that the process was increasing RAM usage continuously, in the end getting probably oomkilled, while not having started the transfer.

Last lines from the terminal: 
```
[2026-08-30][00:15:58][ftp_client_gui_lib::provider_commands][INFO] Uploading via provider: /home/source-dir/large-file.7z -> /home/target-dir/large-file.7z
[2026-08-30][00:15:58][ftp_client_gui_lib::providers::sftp][INFO] providers::sftp: using native rsync delta transport (host key pinned)
[2026-08-30][00:18:08][ftp_client_gui_lib::credential_store][INFO] Credential stored: config_app_settings
[2026-08-30][00:18:10][ftp_client_gui_lib::credential_store][INFO] Credential stored: config_app_settings
[2026-08-30][00:18:10][ftp_client_gui_lib::credential_store][INFO] Credential stored: config_app_settings
[2026-08-30][00:18:11][ftp_client_gui_lib::credential_store][INFO] Credential stored: config_app_settings
[1]    10119 killed     aeroftp
```

btw. a really nice application 👍

axpnet:
Thank you for this, and sorry it took a while to answer here.

[...]

PR #661, opened 2026-08-30
fix(aerorsync): bound memory for huge uploads

axpnet:
## Summary

- stream native AeroSync upload delta generation, compression, encoding, and mux writes in bounded 4 MiB slabs
- remove all production file-sized diagnostic/op/payload accumulation from the streaming path
- replace product `rsync -c` with `-I`, eliminating the full pre-transfer checksum read and rewind while preserving whole-file integrity through the delta trailer
- retain the historical `-c` shape for capture/oracle profiles
- fail fast when a live source grows beyond its declared length instead of waiting indefinitely for EOF
- add causal, protocol-equivalence, one-pass, force-transfer, live-rsync, and manual huge-file acceptance coverage

## Root cause

`send_delta_phase_streaming` read the source in slabs but retained the complete raw op plan, compressed literal blobs, split wire ops, a cloned report, the encoded delta, and the final mux payload before the first network write. An incompressible 25.2 GB file therefore created several file-sized in-memory generations and could be OOM-killed.

The product path also read the complete source once for the file-list checksum, rewound it, and only then began the actual delta pass.

## Implementation

The upload path now keeps only one source slab, codec state, one slab of engine/compressed output, and the pending mux payload. Each generation is written before the next source read. Stateful zstd/deflate encoders preserve stream semantics across slab boundaries, and the rolling producer can flush pending unmatched literals without disturbing its window.

Product commands use measured compact bundles with `-I`. This forces files already selected for transfer even when size and mtime match, while the sender/receiver trailer still verifies the reconstructed whole file. Capture commands remain byte-compatible with the frozen `-c` oracles.

If an active source grows during upload, the driver now rejects the first slab that would exceed the declared length before hashing or encoding it. No END marker or checksum trailer is emitted, so the receiver cannot commit an overlong or partial reconstruction and an append-only source cannot keep the read loop alive indefinitely.

## Acceptance

Physical deterministic incompressible files uploaded through the native driver to stock rsync 3.2.7 and verified with SHA-256 on both ends:

| Input | Time | Max RSS | SHA-256 |
| --- | ---: | ---: | --- |
| 5,000,000,000 bytes | 61.23 s | 90,848 KB | `f48c00f9893e6c6ad85085fdbd128e9e169d9b0a9e28995742ac30ab38ed4136` |
| 25,200,000,000 bytes | 293.94 s | 90,408 KB | `ed7c19fa295a2d731cdb9513a61c7db53c4f11b0744f760f1348ed218c940247` |

The input grew 5.04x while peak RSS remained flat (440 KB lower on the larger run). The incompressible wire-byte lower bound also passed.

`-I` force-transfer was verified live against a remote baseline with identical size and mtime but different content; the resulting remote SHA-256 matched the source.

## Verification

- red-first causal test on exact v4.1.8 tag: fails before the fix because another source slab is read before the first incremental write
- source-growth causal regression: `InvalidFrame` after the first overlong read and no subsequent source poll
- `cargo test --features aerorsync --lib 'aerorsync::'`: 675 passed, 0 failed
- `cargo clippy --features aerorsync --all-targets -- -D warnings`
- live streaming upload against stock rsync: passed, byte-identical
- live native delta download against stock rsync: passed, byte-identical
- `npm run ci:pre-push`: exit 0 (security suite, 826 Vitest tests, typecheck, i18n validation, 3,655 core Rust tests, 509 CLI tests, integration suites)

The gate was green except for the yanked-package verification, which did not run to completion because the crates.io registry request timed out. No RustSec advisory finding was reported; this unavailable check still needs to be distinguished from a pass.

#658


axpnet:
Holding this for its author, who is at a session limit until roughly 19:50 CEST. Nothing here changes the code; it records what the review finding turns out to be, so the fix takes minutes rather than a re-derivation.

**The finding is real and it is more than wasted work.** `declared_last_slab` is an **equality**, `total_source_bytes == source_len` at `native_driver.rs:3024`. If the source grows while it is being read, the counter steps past `source_len` and that equality never holds again. So `producer.finalize()` is never called, the loop keeps reading, hashing, encoding and writing slabs, and the mismatch check at 3071 sits after the loop it can never be reached from. On a source that is still being appended to, the loop does not terminate.

That is a liveness bug inside the change whose whole subject is bounding, which is worth stating plainly rather than filing under minor.

**Reachability**: a file being appended to during upload. A log, an active recording, something another process is still writing. Not everyday, not exotic.

**Smallest correct shape**, and the choice matters: returning `InvalidFrame` as soon as the counter exceeds `source_len` fails fast and says why. Relaxing the comparison to `>=` would also end the loop, but it would end it by treating an overlong source as if it were the declared one, finalising on a boundary the caller never declared, so the wire would carry more than `source_len` and the receiver would be told nothing. Fail, do not accommodate.

**Merge caveat, independent of the above.** The branch's commit message still carries `Fixes #658`, and #658 was opened by an external user. The pull request body has already been corrected. With a squash merge the branch messages do not reach the default branch, so this only fires if the squash body is left as GitHub's default, which concatenates them. **Whoever merges must supply an explicit squash body.** The convention here is that a reporter closes their own issue after confirming on their own setup, and this one was measured against `/dev/urandom` at the reported size, which is the right input and is still not their file, their disk or their link.

The rest of the acceptance is in good shape and worth keeping visible: incompressible input at two sizes, 5.0 GB and 25.2 GB, digests matching on both ends, and peak RSS at 90,848 KB against 90,408 KB for a 5.04x larger input. Boundedness proved by a peak that does not scale, rather than by one run that did not die.

PR #663, opened 2026-08-30
feat(ftp): bound opening a data channel, where the wait actually lives

axpnet:
**The wait was never in the loop.** Measured on `main`, three runs of three against the FTPS lab: a download of a file the server refuses hangs with the control socket holding an unread reply and the data socket showing `bytes_sent 388`, `bytes_received 0`. 388 out and none back is a TLS ClientHello nobody answered, so the wait is inside the data connection's handshake, which happens BEFORE the first read. `read_watching_control` guards the read loop, and the loop is never reached. #660 gave nine sites a guarded loop; this gives them a guarded opening, which is where the defect lives.

**Two unbounded awaits, and both are covered.** In the crate's `data_command`, active mode is bounded by `active_timeout`, but the passive connect (`passive_stream_builder`) and the TLS upgrade (`tls_ctx.connect`) are not. Only the second was measured. The first is the one that matters on a **plaintext** session, where there is no handshake to hang in, and no measurement in this campaign has ever exercised the plaintext path: it is not that there are no defects there, it is that nobody has looked. Bounding only the measured one would have closed the case we had seen and left the case we had not.

**`open_with_cap` bounds and does nothing else, and that is a correction.** The first version classified the failure too. Writing the SECOND caller showed why that was wrong: `STOR` maps with `map_store_error` and the others with `classify_data_failure`, so a helper that classified would have changed that site's behaviour silently. It is the second consumer that reveals an abstraction doing one thing too many; the first cannot, because whatever the helper does is what that caller wanted.

**What happens when the cap expires, and why the session usually survives.** The command is already on the wire and its reply unread: `pasv()` reads its own `227`, but `perform(cmd)` reads nothing. So `after_timed_out_open` looks before it concludes. The `Recv-Q` of 55 during the measured hang decomposes exactly as a TLS 1.2 AES-GCM record carrying `550 Failed to open file.\r\n`: 5 header + 8 nonce + 26 payload + 16 tag. Consistent, not proof, but it says the likely case is that the server refused long ago, which is precisely why it never serviced the data port. So the cap does not produce a quick failure; it produces the server's own reason, which was there to be read.

**Four conditions, and a named silence.** The session is kept only when the reply is parsed WHOLE, is FINAL, belongs HERE, and does not announce the channel closing. That last one is `421`, "service not available, closing control connection": a `4xx`, final, legitimately present, and kept by every earlier version of this condition, while the server was saying the channel was going away.

This arm has now been narrowed four times, and the sequence is the useful part. Keeping on any error was right about "something failed" and silent on whether a whole reply arrived. Keeping on the error VARIANT was right about that and silent on finality. Keeping on the STATUS CLASS was right about finality and silent on the connection surviving. **Each discriminant was correct for the property it was chosen for and mute on the next**, which is why it looked total every time: totality is answered with respect to one property at a time.

So the question that matters for the fifth round is not "does this cover every case" but **"which property is this discriminant silent about"**. This one is silent about WHOSE reply it is: FTP carries no request identifier, so a reply queued here could belong to an earlier command answered late, and consuming it as ours would leave our own still owed. Nothing in reach distinguishes them, so it is named in the code rather than claimed closed.

**And the session is discarded whenever it cannot be handed on**, which is what makes the cap safe:

- a complete reply was parsed: it was consumed whole, the control channel is aligned again, the session survives;
- the collection produced nothing readable: `read_response_in` is not cancellation-safe and may have been dropped mid-line, so the session goes;
- nothing was queued at all: the reply is still owed, and a session one answer behind hands the next question somebody else's answer. The session goes.

The second of those is a CATCH-ALL, not a third named case, and that is deliberate. `BadResponse`, a connection error, and any variant a later version of the crate might add do not share a meaning; they share the absence of one, since none of them establishes that the control channel is aligned. So the default discards, because the safe side is the one that does not assume. It is the same asymmetry as the write-side record types with the sign reversed: there the safe default was to keep writing, here it is to throw the session away, and in both the safe default is the branch that claims to know least.

**Listings get a TOTAL cap, not an inactivity one.** A transfer wants inactivity, because a large file legitimately takes hours and a total cap would kill it. A listing has an intrinsic size, so an inactivity bound would never fire against a server that dribbles a row at a time. The expiry is returned as an `FtpError` so the three listing sites treat it exactly as they already treat a refusal: `MLSD` marks itself broken and falls through to `LIST`, which has its own cap.

**The two numbers, with what they rest on.** `OPEN_BUDGET` is 30s, chosen to sit far from both measured edges: a healthy open against the lab completes in 0.116s and the hang was still unresolved at 120s, which was the probe's cap rather than a limit of the wait. `LIST_BUDGET` is 300s and is **reasoned, not measured**: it is a liveness limit, not a performance one, and it has NOT been checked against a pathological listing such as a hundred thousand entries over a slow link. Raising it is safe; lowering it is not. If a truncated listing is ever reported, this is the first place to look, not the network. It deliberately does not copy the twin's `list_timeout` of 30s in `src/ftp.rs`, which is the number we suspect may already be truncating legitimate listings there: copying it would have inherited the answer without asking the question.

**This bounds ONE opening, and a transfer makes as many as the caller asks for.** Measured against a permanent stall (23), counting openings while varying `--retries`: 0 gives 2, 1 gives 2, 2 gives 4, 3 gives 6, 5 gives 10. So the count is `2 x max(1, retries)`: two openings per attempt, times the attempts the caller chose. The 6 first observed was simply 2 x 3, the default, and quoting it as a constant would have made a caller's setting look like a property of the client.

The user-visible worst case is therefore `2 x retries x OPEN_BUDGET`: about 180s at the default, 600s with `--retries 10`. That multiplier is literally a knob the user turns, which settles where the number belongs: the cap is the only one of the two that does not depend on the caller, so lowering it to make the total tolerable would be answering a retry question with a transport constant, at the wrong layer. It is stated because whoever tunes `OPEN_BUDGET` is not setting one sixth of the worst case, they are setting a fraction that is not fixed, and a budget sized once against a user's patience does not stay sized. Either way it is bounded now, and it was not before: the same download did not finish at all.

**Divergence on the four sites #660 called almost-no-op.** `download_to_bytes`, `resume_download`, `read_range` and `download_single` gain the opening cap here. Anyone who read #660 has "those four are unchanged" in mind, and nobody re-reads a closed pull request, so it is stated rather than left to be deduced. The five that had nothing gain it too, and `ftp_download_one_range`, which had a cancel token and no deadline at all, gains its first bound.

**Out of scope, deliberately.** Driving the `LIST` loop ourselves and giving `STOR` its own surveillance are new behaviour with no inherited tests, and they answer a different question: this pull request asks "does the opening have a limit?", which is mechanical and uniform across nine sites, while that one asks "is the new behaviour right?". They are separated for the same reason #660 was separated from this.

**Gate.** `cargo fmt`; `clippy --all-features --tests` clean; 36 `providers::ftp` tests pass. The live proof needs the stalling fixture, which is #662 and not yet on `main`; the lane that runs it comes with the test, and no code here depends on it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added time limits for FTP file transfers and directory listings.
  * Improved handling of stalled or unresponsive FTP operations, including directory listing and resumed or ranged transfers.
  * FTP sessions are now safely closed when recovery is not possible, reducing hangs and inconsistent connection states.
  * Prevented delayed server responses from interfering with subsequent FTP commands.
  * Improved recovery from failed data-channel connections.

### DIFF
--- a/.github/workflows/ftp-mlsd.yml
+++ b/.github/workflows/ftp-mlsd.yml
@@ -19,6 +19,7 @@ on:
       - 'src-tauri/src/providers/types.rs'
       - 'src-tauri/tests/integration_ftp_mlsd.rs'
       - 'src-tauri/tests/fixtures/ftp-mlsd/**'
+      - 'src-tauri/tests/fixtures/ftp/**'
       - '.github/workflows/ftp-mlsd.yml'
   pull_request:
     branches: [main]
@@ -28,6 +29,7 @@ on:
       - 'src-tauri/src/providers/types.rs'
       - 'src-tauri/tests/integration_ftp_mlsd.rs'
       - 'src-tauri/tests/fixtures/ftp-mlsd/**'
+      - 'src-tauri/tests/fixtures/ftp/**'
       - '.github/workflows/ftp-mlsd.yml'
   workflow_dispatch:
 
@@ -158,3 +160,82 @@ jobs:
         if: always()
         working-directory: src-tauri/tests/fixtures/ftp-mlsd
         run: docker compose down -v
+
+  # A listing that times out must not hand the session on. This runs the guard
+  # for that as its own job rather than inside `mlsd`, because it needs a
+  # different server: `slowftp.py --feat mlsd-hang` accepts MLSD and then says
+  # nothing, which the docker fixture above cannot do.
+  listing-timeout:
+    runs-on: ubuntu-latest
+    # The test itself waits out LIST_BUDGET, which is 300s and not injectable,
+    # and a cold Tauri build precedes it. 90 is the same cap the job above uses
+    # for the same shape of work, and it is not tight: the wait is five minutes
+    # of the ninety.
+    timeout-minutes: 90
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: src-tauri
+
+      # 420s of hanging, not 90: the fixture must outlast the 300s budget under
+      # test. At the old default of 90 the fixture gave up first, and a test that
+      # believed it was measuring the cap measured the fixture and passed.
+      - name: Start the hanging-MLSD fixture
+        working-directory: src-tauri/tests/fixtures/ftp
+        run: |
+          python3 slowftp.py --port 2140 --feat mlsd-hang --hang 420 \
+            > /tmp/slowftp-mlsd-hang.log 2>&1 &
+          echo $! > /tmp/slowftp.pid
+
+      # The banner, not the process: a pid says the interpreter started, not that
+      # it is serving. 20 iterations worst case is 20 x (3s read + 1s sleep) =
+      # 80s, comfortably inside the two-minute cap, so the loop finishes before
+      # the cap and the step cannot die mute.
+      - name: Wait for the fixture to answer
+        timeout-minutes: 2
+        run: |
+          for i in $(seq 1 20); do
+            if exec 3<>/dev/tcp/127.0.0.1/2140 2>/dev/null \
+               && read -r -t 3 -u 3 banner \
+               && [[ "$banner" == 220* ]]; then
+              echo "fixture answering after ${i}s: $banner"
+              exec 3<&- 3>&-
+              exit 0
+            fi
+            exec 3<&- 3>&- 2>/dev/null || true
+            sleep 1
+          done
+          echo "the fixture never answered on 2140"
+          exit 1
+
+      # The count is checked because this selects by NAME with `--lib`, the shape
+      # that prints "test result: ok" and exits 0 when the filter matches
+      # nothing. Exactly one, not "at least one": this step promises to run one
+      # named test, so a second arriving here is a change to what the job does
+      # and should be said out loud. The sibling aerorsync lane uses `-lt`
+      # instead, and correctly: it collects a whole module, which may grow.
+      - name: The session must not survive a timed-out listing
+        timeout-minutes: 40
+        working-directory: src-tauri
+        run: |
+          set -o pipefail
+          cargo test --lib providers::ftp::live_listing_timeout -- \
+            --ignored --nocapture | tee /tmp/listing-timeout.log
+          ran=$(grep -oE '^test result: ok\. [0-9]+ passed' /tmp/listing-timeout.log \
+                | grep -oE '[0-9]+' | head -1 || true)
+          echo "live listing-timeout tests executed: ${ran:-none}"
+          if [ "${ran:-0}" -ne 1 ]; then
+            echo "::error::expected exactly 1 test, ran ${ran:-0}: the filter stopped matching,"
+            echo "::error::which is a green that executed nothing. Fix the filter, not this check."
+            exit 1
+          fi
+
+      - name: Fixture log on failure
+        if: failure()
+        run: cat /tmp/slowftp-mlsd-hang.log || true
+
+      - name: Teardown
+        if: always()
+        run: kill "$(cat /tmp/slowftp.pid)" 2>/dev/null || true
--- a/.github/workflows/ftp-mlsd.yml
+++ b/.github/workflows/ftp-mlsd.yml
@@ -179,6 +179,56 @@ jobs:
         with:
           workspaces: src-tauri
 
+      # The three steps below are copied from the `mlsd` job, deliberately and
+      # in full. This job was added beside that one and inherits nothing from
+      # it: without them `cargo` reaches `glib-sys`, cannot find `glib-2.0`, and
+      # fails with exit 101 before a single test runs, which is what happened on
+      # the first attempt. The fixture had started correctly; the build never
+      # got to it.
+      #
+      # Copying only the install would have been worse than copying nothing,
+      # because the failure would have moved instead of going away: that mirror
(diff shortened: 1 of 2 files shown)


axpnet:
@coderabbitai review

Requesting a look at `6209c29b` specifically. Your two earlier reviews on this branch each found a real defect in the same match arm, and this head is its fourth revision, so it is worth the pass rather than my reading alone.

For context on what changed since `0abc35f0`: `421` is excluded from the retained-session branch, because it is a `4xx` and also means the server is closing the control connection, so keeping the session hands on one the server has just said it is ending. The session is now discarded for it.

The sequence is what makes this arm worth a fourth look. Keeping on **any error** was wrong because a malformed reply can mean half a line consumed. Keeping on the **variant** was wrong because a queued `1xx` parses the same way and its completion then answers the next command. Keeping on the **status class** was wrong because of `421`. Each discriminator was correct about the property it was chosen for and silent about the next one down, which is why every version looked total.

One limitation is named in the comment rather than closed: the current test is mute on **whose** reply it is. FTP carries no request identifier, so a reply queued at that moment could belong to an earlier command whose response arrived late, and nothing available distinguishes them.

axpnet:
Reviewing this branch for the property e4 found in `after_timed_out_open`, I checked whether the sibling cap got the same treatment. It did not, and the reconnect that looks like it covers the gap does not fire.

**`list_with_cap` hands back a desynchronised session, and every caller keeps using it.**

On expiry the helper returns an error and nothing discards the stream. The dropped future was inside `read_response_in`, which this file's own comments state is not cancellation-safe ("dropping it mid-line leaves the control channel misaligned"), and the `MLSD`/`LIST` reply is still owed. All three call sites then reuse that session:

- `mlsd()` at :443 sets `mlsd_broken = true` and **falls through to `LIST` on the same session**, which is reuse by design;
- the `-a` path at :480 calls `restore_cwd` on the same stream before reporting, so the late listing reply is read as the `CWD`'s;
- :505 maps to `ServerError` and the session stays.

**The reconnect guard does not catch it, and the reason is the interesting part.** The fall-through is guarded by `is_stale_data_connection_error`, which dispatches on message SUBSTRINGS: `data connection is already open`, `425` + `data connection`, `invalid syntax`, `invalid response`, `connection reset`, `broken pipe`. The cap's error renders as `Connection error: no listing within 300s` (suppaftp `types.rs:22`, `#[error("Connection error: {0}")]`). It matches none of them, so `reconnect_after_data_error` is never reached.

That is a guard that dispatches on text failing to cover an error introduced afterwards, because the new error's wording was never added to it. Nothing at the cap's own call site shows this, which is why it survived four revisions of the neighbouring arm.

**Severity, measured rather than argued.** A session one reply behind does not resynchronise: the shift moves forward. Independently measured on this campaign, a server sending two replies in one write makes the NEXT command read the previous one's answer, and where that answer is positive the command reports success it never had, silently. Here the owed reply is a listing completion, so the following `LIST` or `CWD` can read a `226`/`250` as its own.

**Suggested shape, not a patch.** Discard the session when the cap expires, exactly as `after_timed_out_open` already does when nothing is queued, and for the reason its own doc gives: "a session one answer behind hands the next question somebody else's answer." The cost is one reconnect on a path that has already spent 300 seconds. `list_with_cap` is a free function over a future, so the discard belongs either at the three call sites or in a variant taking `&mut self`; that is the author's call.

Widening `is_stale_data_connection_error` would also silence it, and it is the worse of the two: it keeps the dispatch-on-text mechanism that produced the gap.

This is not an objection to the cap. Before it, the wait was unbounded, which is worse than a desynchronised session. It is the second half of the same fix.

axpnet:
Reviewed `f27c056c`. The shape is better than what I asked for: returning `Ok(None)` puts the expiry in the TYPE, so the three sites cannot walk past it and the compiler enforces what a text match could not. It also makes `list_with_cap` symmetric with `open_with_cap`, which is what the review question should have produced in the first place. The `-a` site handles the expiry before `restore_cwd`, which is the ordering that matters, and `is_stale_data_connection_error` is left alone with the reason written beside it.

**One thing is lost on the way out, and it is state rather than session.**

The `MLSD` site now returns at `Ok(None)` BEFORE the arm that runs `self.mlsd_broken = true; self.mlsd_supported = false;`. Discarding the session is right, and falling through to `LIST` on that same session would be the defect. But `mlsd_broken` is provider state, not session state: it is a field (`ftp.rs:65`), and the connect path reads it at `ftp.rs:1121` as `self.mlsd_supported = server_supports_mlsd && !self.mlsd_broken`, so it survives a re-dial and suppresses `MLSD` on the next attempt.

Skipping it means a server whose `MLSD` hangs is never learned about. Every listing spends the full 300s budget, fails, discards the session, and the next call tries `MLSD` again. The fallback that exists precisely for this server never engages. Before this branch such a server hung once and for ever, so this is still an improvement, but it leaves the provider permanently slow instead of permanently stuck, and the difference from "fixed" is one assignment.

Suggested: set `mlsd_broken` (and `mlsd_supported`) on the expiry path before returning, so the session is discarded AND the next attempt goes straight to `LIST`.

Worth naming the shape, because it is the same one twice in this file: an early return that is correct about the resource and silently skips the bookkeeping the ordinary error path performs. The first instance was the timeout error that did not match the reconnect list; this one is the timeout return that does not reach the flag. Both are paths added later beside an existing one, and both inherit none of what the existing one does.

PR #664, opened 2026-08-30
fix(transfer): cap what a segmented download holds, not just one request of it

axpnet:
## The cap that never applied where it mattered

`read_range` returns a `Vec`, so each sub-read is materialised whole before it is written. The per-request cap of 64 MiB bounds one of those, and its comment says it stops a single range request growing unbounded on very large files. That is true, and it reads as though it bounded the download.

It does not. The window workers run concurrently, `max_parallel` equals the segment count, and the peak is their product.

The arithmetic is worse below the cap than above it, which is why nobody had met it. A sub-read is `min(window, 64 MiB)`, and a window is the file divided by the segments, so under a gigabyte the sub-read IS the window and every worker holds a sixteenth of the file:

```
    file   segments   window   sub-read   planned peak
  200 MB         16      12M        12M         200 MB
  500 MB         16      31M        31M         500 MB
    1 GB         16      64M        64M           1 GB
   25 GB         16    1600M        64M           1 GB
```

The planned peak is `min(file size, 1 GiB)`. A 200 MB download holds 200 MB of it in memory, and nobody goes looking for a memory defect at that size. The protection starts exactly where the ratio stops being surprising.

## What changes

The aggregate is capped as well as the request. The per-request cap stays for what it was always for.

It is paid for in round trips. On FTP each sub-read is a REST plus RETR on a fresh data connection, with a TLS handshake when the channel is protected, so a smaller sub-read means more of them. **The budget is a starting point, not a tuned value**, and the direction of doubt is written next to it: too small costs speed on every large download, too large costs memory on every concurrent one. It wants measuring on a real link before anyone treats the number as settled.

The existing minimum chunk size floors the division, so dividing a budget among many segments can never ask for a pathologically small request. That is asserted separately.

## The test

The peak is now decided by arithmetic in one place, so it can be checked without running a download.

It uses **two sizes** and asserts the peak does not follow the file. A single figure, however small, is a statement about one run; the question a memory defect asks is whether the number tracks the input.

It did. With the aggregate cap removed, the row reports 200 MB of file planning 200 MB and 25 GB planning 1 GB, which is the defect stated in its own numbers. That was confirmed before the fix was kept.

## What this does not resolve

There are two segmented download engines in this tree with opposite memory policies. `ftp_download_one_range` reads into a fixed buffer and writes with seek plus write_all, so its memory does not depend on the window; the generic path goes through `read_range` and materialises. Both serve FTP, and which one runs depends on the surface the download came from.

Removing that split means changing what `read_range` returns, which is a trait shape with eleven implementations. It is not attempted here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced peak memory usage during segmented downloads by applying an aggregate 128 MiB memory budget across concurrent download segments.
  * Improved download stability for large files while preserving a minimum read size for efficient transfers.
* **Tests**
  * Added coverage verifying memory usage remains bounded and behaves correctly across different file sizes and segment counts.

### DIFF
--- a/src-tauri/src/provider_transfer_executor.rs
+++ b/src-tauri/src/provider_transfer_executor.rs
@@ -99,6 +99,55 @@ const SEGMENTED_DOWNLOAD_MIN_CHUNK_SIZE: u64 = 1024 * 1024;
 /// unbounded on very large files.
 const SEGMENTED_DOWNLOAD_SUB_READ_SIZE: u64 = 64 * 1024 * 1024;
 
+/// Ceiling on the bytes all window workers may hold at once.
+///
+/// `read_range` returns a `Vec`, so a sub-read is materialised whole before it
+/// is written. The per-request cap above bounds ONE of those and says so, and
+/// that was read as bounding the download. It does not: the workers run
+/// concurrently, `max_parallel` equals the segment count, and the peak is their
+/// product.
+///
+/// The arithmetic is worse below the cap than above it. A sub-read is
+/// `min(window, 64 MiB)` and a window is the file divided by the segments, so
+/// under a gigabyte the sub-read IS the window and every worker holds a
+/// sixteenth of the file. A 500 MB download peaks near 500 MB, a 200 MB one
+/// near 200 MB, and only past a gigabyte does the per-request cap begin to
+/// bite. The protection starts exactly where the ratio stops being surprising.
+///
+/// So the aggregate is capped too, and the per-request cap stays for what it
+/// was always for. What this buys is paid in round trips: on FTP each sub-read
+/// is a REST plus RETR on a new data connection, with a TLS handshake when the
+/// channel is protected, so a smaller sub-read means more of them. The number
+/// below is therefore a starting point that wants measuring on a real link, not
+/// a tuned value, and the direction of doubt is stated: too small costs speed
+/// on every large download, too large costs memory on every concurrent one.
+const SEGMENTED_DOWNLOAD_TOTAL_READ_BUDGET: u64 = 128 * 1024 * 1024;
+
+/// Bytes a single `read_range` may ask for, given how many windows are running.
+///
+/// Separated from the loop so the memory a download plans to use can be
+/// checked without running one: the peak is decided here, by arithmetic, and a
+/// test that asserts it does not scale with the file is a statement about the
+/// property rather than about one observed run.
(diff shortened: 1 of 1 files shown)


axpnet:
Reviewed `e0267cb9` against `576b8abc`. The arithmetic checks out: before this, `sub = min(window, 64 MiB)` and `peak = sub * segments`, so the peak was `min(file_size, 16 * 64 MiB)`, that is `min(file_size, 1 GiB)`. The per-request cap could not bind below a gigabyte because there the sub-read IS the window. The stated fact is the real one.

Two things I checked rather than assumed, both fine:

- The anti-fragmentation branch in `provider_segmented_effective_count` cannot raise the count. It runs only when `chunk < MIN_CHUNK`, which implies `file_size < segments * MIN_CHUNK <= 16 MiB`, so `file_size / MIN_CHUNK < 16`. It only ever reduces.
- The floor and the budget do not fight at the sizes reachable today: with the clamp at 16, `share` is 8 MiB, comfortably above the 1 MiB floor.

**One latent coupling worth a line, not a change.**

`segmented_sub_read_size` floors the share at `MIN_CHUNK_SIZE`, and that floor overrides the budget. The aggregate therefore stays under `TOTAL_READ_BUDGET` only while `segments <= 128` (at 128, `share` is exactly 1 MiB and the product is exactly the budget; above it, the floor wins and the peak grows linearly again). Today `segments` is clamped to 16 in a different function, so there is an 8x margin and the ceiling is real.

But nothing at the budget's own site records that its validity depends on a constant elsewhere. Raise that clamp, or let `max_workers` exceed it, and the budget silently stops being a ceiling with no test failing: the existing tests use 16 and 64, and even 64 lands exactly at the budget rather than over it. That is the shape this campaign has been finding all week, a guard that holds for a reason written somewhere the reader of the guard will not look.

Cheapest fix is a sentence in the doc naming the dependency, or a `debug_assert` that the product does not exceed the budget so the coupling fails loudly if the clamp moves. Either is enough; the margin is genuinely comfortable.

**On what the PR declares rather than hides**: stating the price in round trips, and writing the budget as a starting point that wants measuring on a real link rather than a tuned value, is the right shape. So is naming the unresolved inconsistency between the two engines instead of leaving it for a reader to discover, given that removing it means changing what `read_range` returns across eleven implementations. A test proved red with the defect stated in its own numbers ("200 MB of file plans 200 MB") is worth more than the fix, because it is the part that stays.

PR #670, opened 2026-08-31
fix(remote-tools): ask whether the parent is a directory, not whether something is there

axpnet:
This site was found by Kimi's error-text gate, which flagged it as a fifth "mkdir already exists" tolerance. It is the opposite: it is the **best** of the five, because it does not stop at the server's words. It already confirms with a `stat`, and the comment beside it records why (ProFTPD on one hosting answers `550 Permission denied` to `MKD` on a directory that exists). Whoever wrote it had already seen that the string was not enough.

**The defect is that the fallback tests the wrong thing.** The line was `if backend.stat(&acc).await.is_ok() { continue; }`, and `stat` returns a `RemoteEntry` carrying an `is_dir` field that the code discarded. So the question asked was "does something exist here" while the question needed is "is there a DIRECTORY here I can descend into".

A regular file where a parent directory should be answers `Ok`, the walk continues, and the next level's `mkdir` fails naming a path one level BELOW the real obstacle. The user is sent to look at `/data/report/2026` when the thing in the way is `/data/report`.

## What changed

`stat` is now read for `is_dir` rather than for success, in the walk and in a new check placed before it. Both the pre-check and the walk report an obstacle through one function rather than two copied `format!` calls, because the message has to be identical from either place that can emit it.

**And it is fewer round trips, not more.** Today the loop issues N `mkdir` calls that all fail when the tree already exists, which is the ordinary case; one `stat` on the parent returns immediately when it is already a directory. When that `stat` says "exists and is not a directory", it fails at once naming the right level. If a backend cannot answer `stat` for a directory, the behaviour falls back exactly to what it is now, which is the path the original comment exists to protect.

`is_dir` was verified reliable before being relied on: `FtpProvider::stat` uses `MLST` where the server offers it, which reports `type=dir`, and otherwise lists the parent and finds the entry. FTP is the protocol this fallback exists for, and it is where the field is populated.

## Three tests, and only two are meant to go red

Two were proved red by restoring the defect, and their messages state the fault rather than the rule:

- `the obstacle is /data/report, but the error points elsewhere: mkdir /data/report/2026 failed` , which is the defect itself, an error naming a level below the obstacle
- `an existing parent directory needs no mkdir at all` , built so it can only pass if no walk happens: only the parent is seeded and the mock refuses every `mkdir`

**The third stays green with the defect restored, deliberately.** It is the non-regression test for the ProFTPD path the comment cites. A red-first test proves the fix was needed and must fail without it; a non-regression test proves nothing else moved and must pass in both versions. Applying "restore the defect and check it fails" to both kinds would either weaken the first or delete the second.

The mock gained a `mkdir_fails_with` field, modelled on the `stat_fails_with` already there, because the first draft of the test called a helper that did not exist: the mock refused every `mkdir` with a fixed string and could not be given a real server's words.

Gate: fmt at zero, clippy on library and CLI at zero, `3663 passed of 3663 executed; 0 failed; 21 ignored`, which is three more than main and they are the three added here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved remote directory creation when encountering existing paths, symbolic links, or other non-directory entries.
  * Prevented directory creation from incorrectly continuing through paths that cannot be safely traversed.
  * Enhanced error reporting to identify the path level that blocked directory creation.

### DIFF
--- a/src-tauri/src/ai_core/remote_tools.rs
+++ b/src-tauri/src/ai_core/remote_tools.rs
@@ -312,6 +312,39 @@ fn parent_remote_dir(path: &str) -> Option<String> {
     }
 }
 
+/// What a listing lets us conclude about descending into a path.
+///
+/// There are two answers and the missing third is the whole point.
+///
+/// A listing can PROVE a path is a directory: `is_dir` comes from the leading
+/// `d` in the permission string (ftp_listing.rs), and the MLSD path sets it for
+/// `type=dir|cdir|pdir`. Nothing a listing says proves a path is NOT one. A
+/// symlink pointing at a directory reports `is_dir == false` because its
+/// permissions begin with `l`, and MLSD recognises only two spellings of a
+/// symlink while RFC 3659 allows a third, so a symlink can arrive with both
+/// flags false. `is_dir == false` therefore means "not shown to be a
+/// directory", never "shown not to be one".
+///
+/// The rule this encodes: refuse only on positive evidence of the negative,
+/// and treat "not proven a directory" as unknown rather than as refuted. It is
+/// a type and not a comment because the same mistake was made twice in this
+/// one function, once in the walk and once in the pre-check written after the
+/// walk was fixed. Removing an instance leaves the reasoning that produced it.
+/// There is no `NotADirectory` variant because a listing cannot hand one out.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum DescendEvidence {
+    ProvenDirectory,
+    Unknown,
+}
+
+fn descend_evidence(entry: &crate::providers::RemoteEntry) -> DescendEvidence {
+    if entry.is_dir {
+        DescendEvidence::ProvenDirectory
+    } else {
+        DescendEvidence::Unknown
+    }
+}
+
 async fn ensure_remote_parents(
     ctx: &dyn ToolCtx,
     server: &str,
@@ -332,13 +365,18 @@ async fn ensure_remote_parents(
     //
     // If the backend cannot stat a directory, this falls through to the walk,
     // which is exactly the behaviour before this check existed.
-    match backend.stat(&parent).await {
-        Ok(entry) if entry.is_dir => return Ok(()),
-        Ok(_) => return Err(non_directory_in_the_way(&parent)),
-        Err(_) => {}
+    if let Ok(entry) = backend.stat(&parent).await {
+        if descend_evidence(&entry) == DescendEvidence::ProvenDirectory {
+            return Ok(());
+        }
+        // Unknown. Not a reason to stop: fall through to the walk, which is
+        // what happened before this check existed.
     }
 
     let mut acc = String::new();
+    // The first level that exists but is not reported as a directory. It is
+    // almost always the real obstacle, and the failure lands one level below.
+    let mut blocked_at: Option<String> = None;
     let leading_slash = parent.starts_with('/');
     for part in parent.split('/').filter(|p| !p.is_empty()) {
         if leading_slash || !acc.is_empty() {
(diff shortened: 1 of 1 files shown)


axpnet:
**Review, requesting changes.** Posted as a comment because GitHub refuses a formal request-changes review on a PR opened by the same account; read it as blocking, not advisory.

Reading `stat` for `is_dir` instead of for success is right, and naming the obstacle at its own level is a real improvement. One thing needs to change before this lands, and it is in the half that is new rather than the half that was fixed.

## The early exit is positive on success and not on failure

```rust
match backend.stat(&parent).await {
    Ok(entry) if entry.is_dir => return Ok(()),
    Ok(_)                     => return Err(non_directory_in_the_way(&parent)),
    Err(_)                    => {}
}
```

The first arm exits early only on positive proof, exactly as the description says. The second one does not. It treats `is_dir == false` as proof that the path is not a directory, but `is_dir` is not a two-valued answer to that question. For one common entity it means "I have not looked".

## Which brings the symlink parent back

A parent that is a symlink to a directory is reported by both FTP paths with `is_dir == false`:

- `ftp_listing.rs:303`, the LIST parser: `is_dir = permissions.starts_with('d')` and `is_symlink = permissions.starts_with('l')`. A symlink row starts with `l`, so it is not a `d`.
- `ftp_listing.rs:459`, the MLSD/MLST parser: `is_dir` is true only for `type=dir|cdir|pdir`, while a symlink arrives as `type=OS.unix=slink`.

The repo already records this. The listing snapshot at `src/ftp.rs:911` pins an `lrwxrwxrwx ... link -> target` row as `dir=false`.

So for `/var/www/html` or `~/public_html` being a symlink to a directory, which is ordinary on shared hosting and is the same ground the ProFTPD comment comes from:

- before this PR: `mkdir` fails, `stat(&acc).is_ok()` is true, the walk continues, the upload works
- after this PR: the pre-check returns `cannot create parent directories: /var/www/html already exists and is not a directory`, and the upload stops

The path is a perfectly good directory to the server, which resolves the link itself. The walk's inner `match` has the same shape and the same consequence when the pre-check could not answer.

This is the symlink-parent regression that was already caught once in this PR's history. It is back, in the new pre-check rather than in the old walk, which is why fixing it the first time did not keep it away.

## The tests cannot see it

`FakeBackend.stats` maps a path to `(is_dir, size)`. There is no symlink dimension, so the entity that breaks cannot be seeded and the three tests are silent about it rather than wrong. Worth adding to the mock alongside `mkdir_fails_with`, since a case that cannot be expressed cannot be regression-tested later either.

## Suggested shape

Three outcomes, because there are three:

```rust
match backend.stat(&parent).await {
    Ok(e) if e.is_dir => return Ok(()),
    // A symlink is not proof of a non-directory. The server resolves it,
    // so fall through and let the walk decide, which is what happened
    // before this check existed.
    Ok(e) if e.is_symlink => {}
    Ok(_) => return Err(non_directory_in_the_way(&parent)),
    Err(_) => {}
}
```

The rule the description already states, hold the fast exit to positive proof, applied to the failing exit too: refuse only on positive proof of a non-directory, and treat "not proven to be a directory" as unknown rather than as refuted.

One warning about the obvious-looking fix. `RemoteEntry::is_walkable_dir()` exists and is `is_dir && !is_symlink`, and its doc says every recursive walk must gate descent on it. It is the wrong predicate here and would make this stricter, not safer. That rule is about a client walking `list` itself, where a link to `..` creates a cycle. This function does not walk anything client-side, it asks the server to create a path, and the server follows the link. A symlinked parent is a legitimate destination here and a trap there.

## Not a change request

The third test staying green with the defect restored is right, and the reasoning for it is worth keeping in the file. A red-first test and a non-regression test are answering different questions and only the first one has to fail without the fix.

PR #675, opened 2026-08-31
fix(cli): decide the keystore exit code on the error variant, not its words

axpnet:
`classify_keystore_error` decided a CLI exit code by searching `e.to_string()` for phrases, while its own doc comment promised the scheme exists "so cron/CI pipelines can branch on it without parsing the error text". The parsing had not been removed, it had been **moved**: what a pipeline no longer had to do became a duty of whoever writes the error messages, and nobody told them.

## The reason, which only the simulation made visible

Three sites interpolate a string the caller chose. So the exit code was a parameter:

| command | message | exit |
|---|---|---|
| `--merge sideways` | `Invalid merge strategy: sideways` | 99 |
| `--merge decrypt` | `Invalid merge strategy: decrypt` | **6**, an authentication failure |
| `--merge "io error"` | `Invalid merge strategy: io error` | **11**, a disk fault |
| `--merge "vault not ready"` | `Invalid merge strategy: vault not ready` | **5**, vault locked |

Same failure, same cause, four codes chosen by what the user typed. `--format` behaves the same way, and the compression-codec marker read out of a backup file could steer it too, so a crafted file picked the exit code of the tool reading it.

## What the first version of this PR got wrong

It declared two changed codes and changed six. The count of construction sites it rested on was wrong: nineteen, not eight. That number came from a `grep | head -8`, eight printed lines read as eight sites, with the truncation added by the same hand that then trusted the result.

Running the old chain over **all nineteen** messages, instead of over the ones expected to matter, shows the four that were regressions:

| site | message | before | first version | now |
|---|---|---|---|---|
| `:190` | `Backup file too large: …` | 11 | 99 | 99 |
| `:203` | `Backup file exceeds cap after read: …` | 11 | 99 | 99 |
| `:890` `:1207` `:1240` | `IO error: Permission denied` (from the store) | 11 | **99** | 11 |
| `:1186` | `Unknown compression codec: …` | 7 | **99** | 7 |

The four in bold replaced a correct answer with none, which is the opposite of the argument this PR makes.

## The remedy is one floor up

Not a different mapping: `Encryption(String)` was swallowing failures that already had a type. The three store sites now map `CredentialError::Io` to `KeystoreExportError::Io`, so 11 follows from the variant rather than from the words "IO error" surviving a flattening. The codec gets a variant of its own carrying 7.

Re-run over the same nineteen, **two** codes move, and they are the two this change is about: a size-cap refusal is not a disk I/O fault, and 99 replaces a wrong answer with no answer.

## Verified

Both halves of the contract are pinned, and each was checked by breaking it: variant to code in the CLI (`UnsupportedCodec` forced to 99 makes the test fail `left: 99, right: 7`), and failure to variant in the producer (flattening `Io` back into `Encryption` makes it fail on "an IO failure from the store must not be flattened").

Local gate: `cargo fmt` clean, `cargo clippy --all-targets -- -D warnings` zero errors, `cargo test` 4202 passed and 0 failed across 23 test binaries.

Whether anything branches on these codes was checked separately and nothing in this repository does: no branch on 5, 6, 7, 11 or 99 in `scripts/` or `.github/`, none of the 31 `aeroftp-cli` invocations touches the keystore, no cron. With a control, since 23 numeric exit-code comparisons do exist, so the pattern would have been found. Limit: this covers our repositories and this user's crontab, not automation elsewhere.

https://claude.ai/code/session_01NhB1WQp9kb1CXaCHSjsX5h

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Keystore CLI error handling for export, import, and info commands.
  * Commands now return consistent exit codes based on the underlying error type.
  * Serialization and encryption failures return exit code 99, while recognized keystore errors retain their specific codes.
  * Fixed cases where errors with similar messages were previously assigned incorrect exit codes.
  * Improved handling and reporting of unsupported compression formats and storage errors.

---

## Merge record

Merged at zero incomplete check runs, read from the check-runs API on the head commit rather than from this page, where a superseded `cancelled` run reads as red.

CodeRabbit reviewed `d3ce8f1a` through `5762711f` with no actionable comments. The two later commits, `8915f15f` and the head `71df9161`, were NOT re-reviewed: the incremental pass was rate limited. So the part that names the rollback classification so a test can call it directly carries a human reading only.

### DIFF
--- a/src-tauri/src/bin/aeroftp_cli.rs
+++ b/src-tauri/src/bin/aeroftp_cli.rs
@@ -37151,6 +37151,7 @@ fn classify_keystore_error(err: &ftp_client_gui_lib::keystore_export::KeystoreEx
         E::InvalidPassword => 6,       // auth failure
         E::VaultNotReady => 5,         // configuration / vault-locked
         E::UnsupportedVersion(_) => 7, // not-supported / unsupported version
+        E::UnsupportedCodec(_) => 7,   // same family: this build cannot read it
         E::Io(_) => 11,                // I/O
         // Both cover too many distinct causes for one code to say anything
         // useful, so they say nothing rather than something false.
@@ -66633,10 +66634,38 @@ mod tests {
             99
         );
 
-        // Same variant, wording that once mattered and now does not.
+        // Same variant, wording that once mattered and now does not. The old
+        // chain gave 6 to anything containing "decrypt", and three sites
+        // interpolate a string the caller chose, so `--merge decrypt` exited 6
+        // and `--merge "io error"` exited 11: a pipeline branching on the code
+        // was being steered from the command line.
         assert_eq!(
-            classify_keystore_error(&E::Encryption("failed to decrypt blob".into())),
+            classify_keystore_error(&E::Encryption("Invalid merge strategy: decrypt".into())),
             99
+        );
+        assert_eq!(
+            classify_keystore_error(&E::Encryption("Invalid merge strategy: io error".into())),
+            99
+        );
+
+        // The two that must NOT move, and did in the first version of this
+        // change. A permission problem on the vault file arrives as a
+        // `CredentialError::Io` and used to reach 11 because its text happened
+        // to say "IO error"; it now reaches 11 because it is an `Io`.
+        assert_eq!(
+            classify_keystore_error(&E::Io(std::io::Error::new(
+                std::io::ErrorKind::PermissionDenied,
+                "Permission denied (os error 13)"
+            ))),
+            11
+        );
+        // A codec this build cannot read used to get 7 from the words "unknown
+        // compression" appearing in an `Encryption` message, which let the
+        // marker inside the backup file pick the exit code. Same 7, from a
+        // variant.
+        assert_eq!(
+            classify_keystore_error(&E::UnsupportedCodec("brotli".into())),
+            7
         );
     }
     use serde_json::json;
--- a/src-tauri/src/bin/aeroftp_cli.rs
+++ b/src-tauri/src/bin/aeroftp_cli.rs
@@ -36138,7 +36138,7 @@ async fn cmd_keystore_export(
     let metadata = match result {
         Ok(Ok(m)) => m,
         Ok(Err(e)) => {
-            let exit = classify_keystore_error(&e.to_string());
+            let exit = classify_keystore_error(&e);
             print_error(format, &format!("keystore export failed: {e}"), exit);
             return exit;
         }
@@ -36259,7 +36259,7 @@ async fn cmd_keystore_import(
     let outcome = match result {
         Ok(Ok(r)) => r,
         Ok(Err(e)) => {
-            let exit = classify_keystore_error(&e.to_string());
+            let exit = classify_keystore_error(&e);
             print_error(format, &format!("keystore import failed: {e}"), exit);
             return exit;
         }
@@ -36401,7 +36401,7 @@ fn cmd_keystore_info(input: &str, json: bool, format: OutputFormat) -> i32 {
     ) {
         Ok(m) => m,
         Err(e) => {
-            let exit = classify_keystore_error(&e.to_string());
+            let exit = classify_keystore_error(&e);
             print_error(format, &format!("keystore info failed: {e}"), exit);
             return exit;
         }
@@ -37120,22 +37120,41 @@ async fn cmd_aerorsync_probe(
     7
 }
 
-/// Map a keystore_export error message back to a CLI exit code.
-/// Matches the 0/1/2/4/5/6/8/11/99 scheme used elsewhere in
-/// `aeroftp-cli` so cron/CI pipelines can branch on it without
-/// parsing the error text.
-fn classify_keystore_error(msg: &str) -> i32 {
-    let low = msg.to_ascii_lowercase();
-    if low.contains("invalid password") || low.contains("decrypt") {
-        6 // auth failure
-    } else if low.contains("vault not ready") || low.contains("vault unavailable") {
-        5 // configuration / vault-locked
-    } else if low.contains("unsupported file version") || low.contains("unknown compression") {
-        7 // not-supported / unsupported version
-    } else if low.contains("io error") || low.contains("backup file") {
-        11 // I/O
-    } else {
-        99 // unclassified
+/// Map a keystore_export error to a CLI exit code, by VARIANT.
+///
+/// Matches the 0/1/2/4/5/6/8/11/99 scheme used elsewhere in `aeroftp-cli` so
+/// cron/CI pipelines can branch on it without parsing the error text.
+///
+/// That promise used to be made and not kept. The function received
+/// `e.to_string()` and searched it for words, so the parsing a pipeline was
+/// told it could skip had merely MOVED: it became a duty of whoever writes the
+/// error messages, and nobody told them. A reworded message changed an exit
+/// code with no test failing anywhere, because the callers had the typed error
+/// in hand and threw the type away to get a string.
+///
+/// It now takes the error itself. Two mappings are worth naming because the
+/// text was getting them wrong:
+///
+/// `Encryption` covers everything from `zstd init` to `Invalid merge strategy`,
+/// and two of its messages begin "Backup file too large" / "Backup file exceeds
+/// cap after read". Those matched a `contains("backup file")` branch and exited
+/// **11**, telling a script a size-cap refusal was a disk I/O fault. They are
+/// unclassified, so they exit 99: that replaces a wrong answer with no answer,
+/// which is the honest one of the two.
+///
+/// The old 6 branch also matched `contains("decrypt")`. No message this enum can
+/// produce contains that word, so that half never fired. Dead, and it looked
+/// like coverage.
+fn classify_keystore_error(err: &ftp_client_gui_lib::keystore_export::KeystoreExportError) -> i32 {
+    use ftp_client_gui_lib::keystore_export::KeystoreExportError as E;
+    match err {
+        E::InvalidPassword => 6,       // auth failure
+        E::VaultNotReady => 5,         // configuration / vault-locked
+        E::UnsupportedVersion(_) => 7, // not-supported / unsupported version
+        E::Io(_) => 11,                // I/O
+        // Both cover too many distinct causes for one code to say anything
+        // useful, so they say nothing rather than something false.
+        E::Serialization(_) | E::Encryption(_) => 99,
     }
 }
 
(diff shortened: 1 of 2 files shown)


axpnet:
**Review, requesting changes.** Posted as a comment because GitHub refuses a formal request-changes review on a PR opened by the same account; read it as blocking, not advisory.

Deciding the exit code on the variant is the right change and I want it to land. What follows is about the enumeration it rests on, because both of this PR's conclusions are derived from it and it is incomplete.

I did not reason about the old chain, I simulated it: every message this enum can produce, run through the exact `contains` sequence from `main`, rather than checking the cases I expected to matter.

## Six sites change exit code, not two

| site | before | after | declared |
| --- | --- | --- | --- |
| `keystore_export.rs:190` "Backup file too large" | 11 | 99 | yes |
| `keystore_export.rs:203` "Backup file exceeds cap after read" | 11 | 99 | yes |
| `keystore_export.rs:890` `Encryption(e.to_string())` on a `CredentialError` | 11 | 99 | no |
| `keystore_export.rs:1186` "Unknown compression codec: {other}" | 7 | 99 | no |
| `keystore_export.rs:1207` `Encryption(e.to_string())` on a `CredentialError` | 11 | 99 | no |
| `keystore_export.rs:1280` "Import failed at '{account}': {e}" | 11 | 99 | no |

`:1240` belongs to the same family as `:890` and `:1207` for the same reason.

## On four of them the argument reverses

The PR justifies the move to 99 as replacing a wrong answer with no answer, and for the two "Backup file" cases that is exactly right: a size-cap refusal is not a disk fault.

It does not hold for the rest. `:890`, `:1207`, `:1240` and `:1280` wrap a `CredentialError` in `Encryption(e.to_string())`. That type has an `Io` variant whose `Display` is `"IO error: {0}"`, and `list_accounts` really does produce it: it goes through `read_vault`, which calls `std::fs::metadata` with `?`. So a permission denied on the vault file rendered as `"Encryption error: IO error: Permission denied (os error 13)"`, and the old chain classified it 11. Eleven was correct there. It is an I/O fault, reported as an I/O fault. After this change it is 99, so a script that could act on it no longer can.

`:1186` has the same shape. `"Unknown compression codec: gzip"` matched `contains("unknown compression")` and exited 7, not-supported, which for an unknown codec is the right thing to say. It becomes 99.

So on four sites this replaces a right answer with no answer, which is the opposite of the trade the PR describes.

## Suggested remedy, in this PR's own direction

The defect is not the mapping. It is that `Encryption(String)` is swallowing errors that are not encryption failures, which is the same category error the typed exit code was meant to remove, one level down. At the sites that wrap a `CredentialError`, map `CredentialError::Io` to `KeystoreExportError::Io` rather than to `Encryption`. Then 11 survives BY VARIANT, which is precisely what this PR is trying to achieve, instead of surviving by a word in a sentence or not surviving at all.

`:1186` needs a deliberate decision rather than a silent one: either a variant that carries 7, or 99 with the reason written next to it. What does not hold is that it changes without being named.

## The `decrypt` branch was not unreachable

I confirm the literal claim: searched case-insensitively across `keystore_export.rs` and `credential_store.rs`, the word appears only in comments, function names, and the `"decrypting"` progress phase, never in an error message.

But three sites interpolate strings chosen by whoever invokes the command: `"Unknown export mode: {other}"`, `"Invalid merge strategy: {other}"`, `"Unknown compression codec: {other}"`. So `aeroftp keystore import --merge decrypt` produced `"Encryption error: Invalid merge strategy: decrypt"`, and the old chain exited **6**, an authentication failure, for a typo in a merge strategy. An account named `backup file` reaching `:1280` exited 11. Both verified in the simulation.

The conclusion to delete that branch is right. The stated reason is not, and the real one is much stronger: an exit code that pipelines branch on was steerable by a word typed on the command line. That belongs in the commit message, because it is the best argument this change has.

## Please recount the sites

The description says all eight construction sites of `Encryption` were read. There are nineteen in the file (twenty matches less the enum declaration). Raising it not for its own sake: the dead-branch finding and the "only these two change" guarantee are both derived from that enumeration, and eleven sites were not in it.

The test is good and I would keep it as is, including the deliberate red before trusting it. It pins what the function does. What the table above changes is the set of inputs the function will now see for a given real-world failure, which is upstream of it.

axpnet:
**Re-review, one change requested.** Posted as a comment because GitHub refuses a formal request-changes review on a PR opened by the same account. The remedy is right and lands exactly where it should: `from_store_error` keeps 11 by VARIANT, and `UnsupportedCodec` keeps 7. Two things below, and the first is a correction to my own last review.

## My "nineteen" was also wrong: there are 23

I told you the file had nineteen `Encryption` construction sites against the eight in the description. Nineteen is what `grep "KeystoreExportError::Encryption("` returns. The real count is **23**: four sites pass the variant as a function rather than calling it, so they have no `(` after the name and my pattern stepped over them.

```
$ grep -c 'KeystoreExportError::Encryption('  keystore_export.rs   ->  19
$ grep -c 'KeystoreExportError::Encryption'   keystore_export.rs   ->  23
$ grep -n  'KeystoreExportError::Encryption)' keystore_export.rs
  1022  .map_err(KeystoreExportError::Encryption)?;   # derive_key_strong
  1025  .map_err(KeystoreExportError::Encryption)?;   # encrypt_aes_gcm
  1151  .map_err(KeystoreExportError::Encryption)?;   # derive_key_strong
  1168  .map_err(KeystoreExportError::Encryption)?;   # derive_key
```

Same class as the `head -8`, different mechanism: a pattern that answers a narrower question than the one being asked, and returns a number that looks like an answer either way.

Those four change nothing, and that was worth checking rather than assuming: they carry `Argon2 params:`, `Argon2 derive:` and `AES-GCM encrypt:`, none of which matches an old needle. I did expect one, because `crypto.rs` renders a decryption failure as `AES-GCM decrypt: {e}` and that string contains `decrypt`, which the old chain read as exit 6. It never reaches this enum: the only `decrypt_aes_gcm` call here discards its error into `InvalidPassword`, which is 6 before and after. So the old `contains("decrypt")` branch, on top of everything else, was one line away from being right by accident.

## `:1300` is a third changing case, and the reason given does not hold for it

You considered it and chose to leave it because it stays 99 either way. That is true for three of the four inner variants and false for the one that motivated `from_store_error`:

| inner `CredentialError` | old | new |
| --- | --- | --- |
| `Io` | **11** | **99** |
| `InvalidMasterPassword` | 99 | 99 |
| `VaultNotInitialized` | 99 | 99 |
| `Serialization` | 99 | 99 |

`Import failed at '{account}': {e}. {n} entries rolled back.` renders an inner `Io` as `IO error: ...`, and the old chain matched `io error` inside the composite. The site wraps `store.store(...)` and `store.delete(...)` during rollback, both of which write the vault, so `Io` is reachable there for the same reasons it is reachable at the three sites you did fix.

It cannot just call `from_store_error`, because that would throw away the account name and the rollback count, which are the useful part of the message. Keeping both:

```rust
Err(e) => {
    let text = format!("Import failed at '{account}': {e}. {} entries rolled back.", committed.len());
    return Err(match e {
        CredentialError::Io(io) => KeystoreExportError::Io(std::io::Error::new(io.kind(), text)),
        _ => KeystoreExportError::Encryption(text),
    });
}
```

That keeps the sentence and the exit code, and it makes the count of changing cases two again rather than three.

## On the two tests

Pinning both halves and saying explicitly that neither test catches a mutation in the other half is the right call, and it is the part I would keep verbatim. A test that pins the classifier and a test that pins the producer are two claims, and a reader who assumes one implies the other stops looking at the half that is not covered.

The `4201 -> 4202` cross-check is good for the same reason the `3660 -> 3661` one was: a total that did not move would have meant the suite was not running what you thought. Worth keeping the habit of stating which single test the `+1` is.

axpnet:
**Re-review of 8915f15f, one change requested.** Posted as a comment because GitHub refuses a formal request-changes review on a PR opened by the same account.

The fix at the rollback site is right, and the comment beside it says the two things a later reader needs: why neither `from_store_error` alone nor `Encryption` alone would do, and why `Io` is reachable there. I re-ran the classification across the store's variants for all four sites and confirm the count: two cases change, and they are the two declared.

## The test does not touch the code it defends

`the_rollback_message_keeps_its_class_and_its_words` calls no production function. It builds the error by hand:

```rust
let text = format!("Import failed at 'acct': {io}. 3 entries rolled back.");
let err = KeystoreExportError::Io(std::io::Error::new(io.kind(), text));
assert!(matches!(err, KeystoreExportError::Io(_)), ...);
```

That is the same construction the production site performs, rewritten in the test, and then asserted. Revert lines 1314-1319 to `Encryption(text)` and this test still passes, because it never reaches them. It is green for the same reason it would be green if the fix were absent.

This is the shape you named yourself on the previous round, that a hand-built error only asserts the mapping back, and it is the same class as the hand-chosen inner value: the check is one step away from the thing it is checking, and the step is invisible because the code in the test looks exactly like the code in the file.

## Suggested shape

Give the classification a name, so the site and the test are the same code:

```rust
fn rollback_failure(
    account: &str,
    e: crate::credential_store::CredentialError,
    rolled_back: usize,
) -> KeystoreExportError {
    let text = format!("Import failed at '{account}': {e}. {rolled_back} entries rolled back.");
    match e {
        crate::credential_store::CredentialError::Io(io) => {
            KeystoreExportError::Io(std::io::Error::new(io.kind(), text))
        }
        _ => KeystoreExportError::Encryption(text),
    }
}
```

The site becomes one call, and the test calls `rollback_failure` instead of rebuilding it. Then reverting the fix turns the test red, which is the property the test is supposed to have.

While you are there, pin the other half too. The test only exercises `Io`; a change that classified EVERY inner error as `Io` would keep it green. One extra assertion on a non-`Io` variant, that it stays `Encryption` and therefore 99, closes that side and costs two lines.

## Worth keeping as written

The doc comment recording that the case was missed because the inner value was invented rather than enumerated is the most useful thing in this diff. "Enumerating the site is not enough if what goes into it is chosen by hand" is the rule, and having it beside the test means the next person to extend the simulation reads it before repeating it.

axpnet:
@coderabbitai review

axpnet:
**Re-review of 71df9161: my objection is resolved, nothing blocking from me.**

Both points are closed, checked on the tree rather than on the description:

- `rollback_failure` exists (`keystore_export.rs:334`), the site is a single call to it (`:1334`), and the test calls the same function (`:1668`, `:1692`) instead of rebuilding the error. Breaking the classification inside the function now turns the test red, which is the property the previous version did not have.
- The non-`Io` half is pinned: `VaultNotInitialized` is asserted to stay `Encryption`, so a change that classified every inner error as `Io` no longer passes, and the rollback count is asserted on that path too.

For the record, the earlier rounds are settled as well: `from_store_error` keeps 11 by variant, `UnsupportedCodec` keeps 7, and the classification across the store's variants for all four sites gives two changing cases, which are the two declared.

The doc comment explaining why this is a named function rather than an inline match is the part worth keeping. It records that the first test rebuilt the error by hand and therefore stayed green with the defect reinstated, and it sits next to the code that a future edit would break, which is where a warning has a chance of being read.

This is my last pass on this PR; I am being stood down. Anything further is for whoever picks it up.

PR #679, opened 2026-08-31
fix(sync): let the classifier see the status shape the providers produce

axpnet:
## What was wrong

Six needles in `classify_sync_error` look for a status code followed by a space: `552 `, `403 `, `550 `, `404 `, `401 `, `530 `. Every failure `providers::ftp` classifies is rendered as `[552] Exceeded storage allocation (while writing ...)`, where the digits are followed by `]`, never by a space. None of those six could fire on a real message. Quota, permission, not-found and authentication were decided by rules that could not see the shape the system produces.

Measured on production-shaped strings before the change: `[552] Exceeded storage allocation (...)`, `[530] Not logged in (...)` and `[502] Command not implemented (...)` all classified as `Unknown`, `retryable = true`. An FTP 5xx is by definition a permanent negative reply, so each was attempted three times.

They were not uniformly dead, which is worse than dead. A server that echoes its own code inside the text, `[550] 550 Failed to open file.`, made the old needle match by accident, so the classification held against one server and silently stopped against another. A defect that always fails is found on the first run because no observation supports it; one that works sometimes produces a TRUE observation supporting a false conclusion, and the evidence in its favour is genuine.

## This changes behaviour, for the majority of servers

Stated up front rather than left to be found, because "revive a dead rule" reads like a no-op and a reviewer has no reason to go looking for which cases move.

| Message | Before | After |
|---|---|---|
| `[550] 550 Failed to open file.` (server echoes) | PermissionDenied, not retryable | unchanged |
| `[550] Failed to open file.` (server does not echo) | Unknown, RETRYABLE | PermissionDenied, not retryable |

For a permanent negative reply that is the right direction: the second attempt cannot succeed. It is still a change, and the population it moves is the larger one.

## How

`mentions_ftp_status` learns the bracketed shape, and the six branches go through a helper that keeps their existing loose search AND adds the anchored bracket.

Anchored, never searched for. The message carries a path the user chose, and a file named `[404] note.txt` must not be read as a status. `classification_text` already removes the paths it can recognise by shape and documents the case it cannot, a relative path with a space matching no `file_path`; the anchoring covers that gap instead of depending on it. That is measured, not assumed: with a whole-message search instead of anchoring, the hostile-path test fails.

Additive on purpose. No message that matched before stops matching. These branches all produce non-retryable kinds, so a message that stops matching falls through to `Unknown`, which is retryable, and a permanent refusal starts being retried again. Tightening the loose half is a separate question, and only the second one is answered here.

## Tests

Three new ones, all starting from messages rendered through `ProviderError` rather than chosen by hand. That matters here: the rule that was dead had a test of its own calling `classify_sync_error("552 Insufficient storage space")`, a string production never generates, so it proved the rule worked on the input the rule was written for.

- the bracketed status reaches the branches: quota and auth, on rendered messages;
- the population that moves: `[550] Failed to open file.` with no echo, asserted on the OUTCOME, since a corpus made only of echoing messages would pass without touching the cases this fix serves;
- the echoing message must now match by construction rather than by accident, so a later simplification of the anchoring cannot keep the test green on the old accidental path;
- a path that looks like a status does not decide the classification.

Both outcome tests were run against the unfixed code and observed failing: `Unknown` against `QuotaExceeded`, `Unknown` against `PermissionDenied`. The anchoring test was run without the anchoring and observed failing too.

Gates: `cargo fmt --all -- --check` clean, `cargo clippy --all-targets -- -D warnings` clean, `cargo test --lib` 3681 passed, 0 failed.

Found while reviewing #676, which is what sends non-path 5xx into this classifier in the first place. Kept out of that pull request on purpose: a narrowing and a change to six classification rules in one diff would ask the reviewer two questions at once.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved FTP error detection across servers that format status codes with or without brackets.
  - More accurately identifies quota, permission, missing-file, authentication, and related transfer errors.
  - Prevents numeric text in file paths or server messages from being misclassified as status codes.
  - Preserves existing retry behavior and error handling for affected FTP operations.

---

## Merge record

Merged at zero incomplete check runs, read from the check-runs API on the head commit rather than from this page, where a superseded `cancelled` run reads as red.

CodeRabbit reviewed through `49779fa5` with no actionable comments. The head commit `f2392578` is documentation only and was not re-reviewed, the incremental pass being rate limited.

What this deliberately does NOT fix, stated so nobody reads the merge as closure: the loose `contains("404 ")` arm is untouched, and with it the case where `connection reset by peer (while writing 404 note.txt)` still classifies as a permanent path failure because of what the file is called. Widening the path detector is a design decision with its own risk, since a "contains a slash" rule would swallow `i/o error`, so it stays open rather than being smuggled into this change.

### DIFF
--- a/src-tauri/src/sync.rs
+++ b/src-tauri/src/sync.rs
@@ -5833,6 +5833,20 @@ mod tests {
     /// `file_path` the caller passed. This asserts the anchoring covers that
     /// gap rather than depending on it, and it was run against a version
     /// without the anchoring, where it fails.
+    ///
+    /// HALF of that gap, and this test must not be read as covering the other.
+    /// Anchoring makes the BRACKETED shape safe, which is the one this change
+    /// introduces. A bare relative name still reaches the loose `contains`
+    /// half, which is older than this change and untouched by it: measured on
+    /// main, `connection reset by peer (while writing 404 note.txt)` comes out
+    /// `PathNotFound`, NOT retryable, while the same failure with `[404]` or
+    /// with `/srv/ok.txt` stays `Network` and retryable. So a connection reset
+    /// stops being retried because of what a file is called.
+    ///
+    /// It is out of scope here for the reason stated on `mentions_status`:
+    /// tightening the loose half is a separate question from letting the rules
+    /// see the shape the system produces, and doing both in one diff would ask
+    /// a reviewer two questions at once. It is tracked on its own.
     #[test]
     fn a_path_that_looks_like_a_status_does_not_decide_the_classification() {
         let raw = "Transfer failed: connection reset by peer (while writing [404] note.txt)";
--- a/src-tauri/src/sync.rs
+++ b/src-tauri/src/sync.rs
@@ -3109,7 +3109,7 @@ pub struct SyncErrorInfo {
 /// unwrapped, because asserting an expected value here only measures what we
 /// already believed.
 fn mentions_ftp_status(lowered: &str, code_and_space: &str) -> bool {
-    if lowered.starts_with(code_and_space) {
+    if opens_with_status(lowered, code_and_space) {
         return true;
     }
     // Lowercased forms of the `#[error("...: {0}")]` labels in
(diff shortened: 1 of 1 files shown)


axpnet:
@coderabbitai review

PR #704, opened 2026-09-01
feat(ai-routing): wire the rule's fallback model, and let Ollama be a routing target

axpnet:
## What this does

Two changes in the AeroAgent Auto-Routing area, one commit each.

**`fix(ai-routing)`: Ollama could not be picked as a routing target.** The Auto-Routing dropdowns listed models from providers matching `isEnabled && apiKey`. Ollama has no API key by design, so the one provider that runs locally, free and fully offline was the only one that could never be selected as the preferred model for a task type. Every other Ollama check in the same panel already reads `apiKey || type === 'ollama'`, including the provider status dot, the key field and the fetch-models button. This filter was the only one that forgot it, and the chat model selector was never affected, which is why the models showed up in the header and not in Settings.

**`feat(ai-routing)`: the rule's fallback model now answers.** `AutoRoutingRule.fallbackModelId` was declared in the settings type and read by nothing, so a rule could name a fallback that never did anything. It is now offered in a second dropdown per task type and it answers when the preferred model fails.

## When the fallback fires, and when it deliberately does not

| Situation | Behaviour |
|---|---|
| 429, 5xx, timeout, network, overloaded, rejected key, unknown model | falls over to the rule's fallback |
| Monthly budget exhausted | never, that is a ceiling the user set on purpose |
| Cancelled request | never, the user stopped it |
| A token has already streamed | never, a partial answer is not replaced with a different voice |
| Model pinned in the chat header | no fallback at all, an explicit choice is not silently swapped |
| Fallback equal to the preferred model | dropped, it would retry the endpoint that just failed |
| Preferred model no longer resolves | its fallback is promoted ahead of the global default |

When the fallback answers, the message signature says so and names the model that failed.

## The part worth reviewing first

The streaming path needed a fix for any of this to be reachable. `ai_chat_stream` reported provider failures through its own catch: it wrote the formatted message into the placeholder and resolved the stream, so the failure never rejected and no caller could react to it. It now rethrows when nothing reached the user, and behaves exactly as before once tokens have arrived.

The `AIChat.tsx` diff looks large because the send body is now wrapped in a retryable function, so almost every line is re-indented. **Read it with `git diff -w`**, which reduces it to 50 lines added and 45 removed.

## Also

The Auto-Routing tip claimed that leaving a task on "Auto (Default)" uses the model selected in the chat header. Routing only runs when no model is pinned there, so what you actually get is the default model. Corrected, and each locale reuses its own translation of that option label rather than the English string.

## Verified

`tsc --noEmit` clean, 98 test files and 858 tests green including 10 new ones covering the resolution and failover rules, `i18n:validate` at 46 locales and 100 percent with 0 placeholders and 0 orphans, production build ok.

## Summary by CodeRabbit

* **New Features**
  * Added fallback-model selection to automatic AI routing, including providers that don’t require an API key.
  * Preserves valid fallback settings when preferred models change.
  * Automatically retries eligible failures with a configured fallback before response streaming begins.
  * Cancels timed-out streams before failover and identifies fallback responses with a badge and tooltip.
* **Documentation**
  * Updated automatic-routing guidance and fallback status messages across supported languages.
* **Tests**
  * Added coverage for routing resolution, failure classification, and failover scenarios.


axpnet:
Thanks, this caught two real defects. Both fixed in `991a6e06`.

**Status-aware classification.** You are right in both directions, and I took the diagnosis rather than just the two examples. A status code now decides when the message carries one, because the status is the fact and the prose around it is not; the phrase list is only the fallback for messages without a status, and it matches on word boundaries. `HTTP 404 Not Found` now fails over, `max_tokens must be <= 5000` no longer does, and `HTTP 400 Bad Request` still does not. Four regression tests cover both directions.

**Cancelling the abandoned stream.** Correct, and worth noting that this leak is one the failover created: before the change the turn ended at the timeout, so nothing kept billing. `ai_cancel_stream` is now called on the timeout path before anything else starts a stream.

**Translations.** Accepted five: Welsh `Dewisol` (optional, not preferred) is now `a ffefrir`; Korean `기본 선택` collided with `autoDefault` and is now `선호 모델`; Hungarian `nem válaszolt` covered only a missing answer and is now `nem volt elérhető`; the Croatian and Polish tips named no second model.

**Not taken, with the reason.** Two comments anchor on `hu.json` `styleLearning` and `hy.json` `stopGeneration`. Those values are unchanged by this branch: they appear in the diff only because a trailing comma was added when the new keys were appended after them. The observations may well be right, but they are pre-existing strings unrelated to routing, so fixing them here would put unrelated locale edits inside a feature branch. They are being logged separately instead.

On `en.json`, I am leaving the tip as it is. Stating that the fallback requires a provider-related failure is accurate but belongs to the contract, not to a one-line hint under a dropdown, and the excluded cases already announce themselves: a budget block shows the budget message, a cancelled request shows nothing. Adding the qualifier would also mean re-translating the tip across 47 locales for a distinction the user never has to act on.

axpnet (commented):
Read on head `991a6e06a` (branch and PR head agree, CodeRabbit's last pass is on this same sha). The routing resolution, the failover gate, the stream rethrow and the cancel-before-failover all read correctly to me, and the ten tests pin the properties rather than the wording. One finding, and it is minor.

**`isFailoverWorthy`: the abort guard does not match `AbortError`.** `/\babort(?:ed)?\b/` needs a word boundary after `abort` or `aborted`, and in `aborterror` the next character is a letter, so the guard never fires on that input. The test `'AbortError'` still passes, but for another reason: nothing else in the message matches either, so the function falls through to `false`. A message such as `AbortError: network request aborted` would match `\bnetwork\b` and fail over on a request the user stopped. Suggest `/\babort/` (no closing boundary), which covers `abort`, `aborted`, `aborting` and `AbortError` alike, and a second assertion on the `'AbortError: network'` shape so the guard, not the fall-through, is what the test proves.

The three CodeRabbit threads still open on this head (en.json wording on the fallback, hu `styleLearning`, hy `stopGeneration`) are wording only, and the last two are strings this PR touches with a trailing comma rather than a change, so they are yours to take or leave.

Not a blocker on merge from my side.

axpnet:
Follow-up on head `7e29f2c1`, answering the two design questions raised after the guard fix, so the answer lives where the merge decision is.

**Broad `abort` guard vs "connection aborted by peer".** The conservative side is right, but the guard answers a narrower question than the substring asks: "did the user stop this". The strings the app produces for that are enumerable (the DOM `AbortError` on an aborted fetch, and whatever the cancel command surfaces), while a transport abort is a different sentence and is exactly the case where another provider is a real remedy. Anchoring the guard on the error-name shape (`/\b(abort|cancell?ed)error\b/` plus the app's own cancel message) and letting "aborted by peer" fall through to the network marker would keep both properties. If the app's cancel message is not enumerable, the broad guard is the right fallback and the rewritten tests are what keep it honest.

**Status-first classification.** Keep it, with one exception: 400. OpenAI and Groq answer an unknown model with 404 `model_not_found`, but several OpenAI-compatible endpoints answer 400 with "model does not exist" or "model not found" in the body. With 400 outside the failover set and status deciding alone, that message never fails over, while another provider may well have the model. So: the status decides for 401, 403, 429 and every 5xx, and a 400 whose body carries a model-not-found marker is failover-worthy. Stated as a known shape, not a verified provider list; a fixture test with both bodies would pin it either way.

Neither is a blocker.

axpnet:
@coderabbitai full review

PR #712, opened 2026-09-01
fix(jottacloud): try both refresh chains and clear a refused one fully

axpnet:
## What this fixes

One station can hold two Jottacloud refresh chains for one profile. A provider bound to a profile id (the GUI) reads and rotates `jottacloud_refresh_<id>`; the CLI never bound the id for a saved profile, so it read and rotated the legacy singleton `jottacloud_refresh`. Jotta rotates each chain independently, so one can be refused while the other still works, and the provider only ever tried the first key it found: a refused per-profile chain fell straight through to the one-shot login token, which is usually spent, and the user saw "Login token expired or already used" on a station that could still connect through the other key.

Measured on 2026-09-01 on the test account: the per-profile chain was refused with `invalid_grant: Session doesn't have required client`, the singleton was accepted by the installed CLI in the same minute, and a profile export took the refused chain because it prefers the per-profile key, so the profile arrived dead on the other station.

## The change

- **Both chains are tried, most specific first.** A bound provider tries `jottacloud_refresh_<id>` then the singleton; an unbound one only the singleton.
- **A refused chain is removed from both places it lives.** The provider used to `store.delete` the raw vault key only, while `resolve_active_credential` reads the active user's partition mirror FIRST, and `mirror_active_credential` had put the chain there (it takes the credential type as an argument, so the key is mirrored without matching the automatic dual writer's `server_*` / `ai_apikey_*` prefixes). Deleted, then found again on the next read: that is how three profile exports in a row carried the same refused chain. The refusal path now deletes the vault key and unmirrors the partition copy, through the primitives that already existed in `user_partitions.rs`.
- **A rotation is written to both keys** when the live chain came from a key other than the provider's own, so an older CLI still reading the singleton finds the current value instead of a consumed one. One profile, one chain, two names while both clients exist.
- **The CLI binds the profile id for Jottacloud** on `--profile`, the key the GUI already uses, so new installs converge on the per-profile key.
- The old loader with its one-shot singleton migration is gone: the dual write supersedes it.

## Verified

- Unit tests on the two pure decisions: candidate order for bound and unbound providers, and the set of keys a rotation is written to in each case.
- A source pin on the property rather than the instance: every file that calls `mirror_active_credential` (exactly `providers/oauth2.rs` and `providers/jottacloud.rs`) must also call `unmirror_active_credential`. MEGA and 4shared store and delete their keys raw and never mirror, so they are outside the property, which an audit of raw `store.delete` sites alone had wrongly flagged.
- `cargo fmt`, `cargo clippy --lib --bin aeroftp-cli --tests -D warnings`, `cargo test --lib providers::jottacloud`, all clean.
- Live, on the test account, with this provider code: one bound connect found the stale per-profile chain, was refused, removed it from both vault and partition, took the singleton, and wrote the rotated token to both keys; the purge live test ran on that connection and passed. The profile was then exported as the last touch on this station (22:22:55Z), imported and connected on the second station (22:23:43Z, which rotates), and this station still connected afterwards (22:24:07Z). Three exports before the fix had carried the same refused chain; this one carried the live one. That measurement also settles a question the failures had raised: one Jottacloud refresh chain serves two stations, a rotation on one does not invalidate the copy on the other.

## Not in this PR

The profile export still takes the per-profile key when present and the singleton otherwise; with the dual write both hold the same value on a station that has connected since this change, so the choice stops mattering there. A station that has not connected since keeps whichever chain it last rotated.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Jottacloud account reconnection by consistently using the correct profile-specific refresh credentials.
  * Added fallback support for legacy Jottacloud credentials.
  * Improved cleanup of invalid or expired refresh credentials.
  * Correctly excludes files marked as deleted when displaying Jottacloud contents.
* **Reliability**
  * Refresh-token updates are now synchronized across supported credential stores, reducing failed reconnections.

### DIFF
--- a/src-tauri/src/bin/aeroftp_cli.rs
+++ b/src-tauri/src/bin/aeroftp_cli.rs
@@ -25937,6 +25937,21 @@ fn profile_value_to_provider_config(
 
     // Load provider-specific options from profile
     apply_profile_options(&mut extra, profile);
+    // Bind Jottacloud to its per-profile refresh chain, the key the GUI
+    // writes. Unbound, the CLI rotated the legacy singleton while the GUI
+    // rotated `jottacloud_refresh_<id>`: two live chains on one station,
+    // each able to die alone, and a profile export that carried whichever
+    // one it found first. One profile, one chain, on every surface.
+    if protocol == "jottacloud" {
+        if let Some(id) = profile
+            .get("id")
+            .and_then(|v| v.as_str())
+            .map(str::trim)
+            .filter(|s| !s.is_empty())
+        {
+            extra.insert("profile_id".to_string(), id.to_string());
+        }
+    }
     if let Some(provider_id) = profile
         .get("providerId")
         .and_then(|v| v.as_str())
--- a/src-tauri/src/providers/jottacloud.rs
+++ b/src-tauri/src/providers/jottacloud.rs
@@ -137,6 +137,11 @@ pub struct JottacloudProvider {
     refresh_token: SecretString,
     token_endpoint: String,
     token_expiry: Instant,
+    /// Vault key the live refresh chain was loaded from, when it was not the
+    /// key this provider is bound to. A rotation is written back to BOTH, so
+    /// whichever client reads either key next finds the current value; see
+    /// `refresh_persist_accounts`.
+    refresh_source_account: Option<String>,
     current_path: String,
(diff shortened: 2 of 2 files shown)


axpnet:
@coderabbitai full review

PR #713, opened 2026-09-01
docs(testing): the single smoke entrypoint exists, say so and point at it

axpnet:
## What

`docs/TESTING.md` still described the single pre-release smoke entrypoint as work in progress ("we are expanding this in two directions: a single pre-release smoke entrypoint..."), while `npm run smoke` (`scripts/smoke.mjs`) has been the pre-tag step of `docs/RELEASE.md` since v4.1.4. The running section now leads with that command and explains the matrix: deterministic suites always run, lab-backed lanes report SKIP when their enabler (Docker fixture port, vault credentials) is absent, and only a real FAIL makes the exit code non-zero.

Alongside, the "frontend tests" command in the same section was `npm run test`, which runs `cargo test`; the frontend suite is `npm run test:unit`.

## Why

Discussion #347, the smoke entrypoint request: the feature shipped but the public testing page kept promising it, so a reader had no way to know it existed. Documentation only, no code change.

### DIFF
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -94,14 +94,22 @@ Legend:
 
 ## Running the tests
 
-Deterministic tests (no credentials, run anywhere):
+One command runs everything that can run on the current machine and prints a pass/skip matrix:
+
+```bash
+npm run smoke
+```
+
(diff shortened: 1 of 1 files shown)


axpnet:
Merge note: CodeRabbit did not review this head (53f2cc009); its only answer was the OSS quota limit at 22:37 UTC. Documentation only, 11 lines in docs/TESTING.md, and the three claims were checked against main before merging: `npm run smoke` maps to scripts/smoke.mjs in package.json, `npm run test:unit` is the vitest suite while `npm run test` is cargo test, and docs/RELEASE.md carries `npm run smoke` as the pre-tag step.

PR #722, opened 2026-09-04
B2: measure the server-reported digest before SSE-B2 lands; R2: reach the eu and us jurisdictions

axpnet:
Two items from the triage of the 2026-09-04 provider and competitor digests, the two that were actionable. The rest of the triage closed items rather than opening work: the Nextcloud 35 RC3 fixes do not touch the path we use, GCS hierarchical namespace does not exist on the S3-compatible endpoint we speak, the S3 403 policy ARN item belongs to the structured-errors track, and the WinSCP 6.6.3 RC was filed as a security-parity priority when this tree already carries OpenSSL 3.6.3 against the 3.5.8 in that release, with the listed CVEs being PuTTY's own C code.

## 1. Backblaze: measure the digest before the service starts encrypting

From **2026-09-14** Backblaze encrypts new uploads and server-side copies with SSE-B2 by default. Nothing changes on the wire from our side, so the only observable is the digest the service reports for an object, which is what `sync`, `dedupe` and `verify` read instead of downloading the bytes.

The existing B2 suite could not have observed it. Its upload tests compare a SHA-256 computed locally before the upload with one computed locally after the download: that proves the bytes survive and says nothing about what the server publishes about them, so a digest that stopped being reported would leave every one of them green while sync silently degraded to size and mtime.

This adds the assertions that can fail:

- single-part upload: `contentSha1` must equal the content SHA-1;
- large file: no digest, which is the documented `B2_UNVERIFIED` behaviour, and if one ever appears it must be the real digest and not an opaque value that passes the 40-hex shape check;
- rename: a server-side copy must preserve it, since the destination bytes are written by the service rather than uploaded with a client-computed SHA-1;
- new `integration_b2_s3_compat`: the S3-compatible endpoint is a **second code path** with no live coverage, and a different exposure, because it reads the S3 ETag and `etag_to_md5` accepts only 32 hex characters, so an opaque ETag makes the digest disappear quietly. AWS keeps the ETag equal to the MD5 under SSE-S3 and only breaks it under SSE-KMS and SSE-C.

### Measured, and the answer is that nothing changes

The wait was avoidable: a bucket can be created with SSE-B2 as its default today, which is the same state Backblaze will switch everyone into. Two throw-away buckets were created, one with default encryption off and one with it on, and the suite was run against both.

| case | encryption off | encryption on |
|---|---|---|
| native, single-part | `sha1=ecfc8e86...23be` | `sha1=ecfc8e86...23be` |
| native, server-side copy | `sha1=bcb72176...4fea` | `sha1=bcb72176...4fea` |
| native, 250 MB multipart | none, as documented | none |
| S3-compat, single-part ETag | `md5=c35cc7d8...f372` | `md5=c35cc7d8...f372` |
| S3-compat, server-side copy | `md5=1910b453...bb77` | `md5=1910b453...bb77` |

SSE-B2 leaves both digests untouched, on both code paths, which is the AWS behaviour for SSE-S3 and not something that could be assumed for B2. So 2026-09-14 needs no client change. The tests stay: they are what makes that a measurement rather than a belief, and what will fail if the service changes its mind later.

Declared limit: the 250 MB multipart case was measured only on the encrypted bucket, to stay inside the free tier's 1 GB daily download cap. It leans away from the conclusion rather than toward it: an unencrypted bucket reporting a digest there would have made the encrypted "none" a regression, and both the checksum matrix and the provider code say a B2 large file never carries a usable `contentSha1`. One transient upload failure was seen on the first multipart run and did not reproduce; its message was lost to an output filter, so it is recorded here rather than explained.

Both files also print what the server actually reported, on a `[b2-digest-baseline]` line, because the comparison that settles this is between a run from **before** 2026-09-14 and one from after, and objects written before the change cannot be created after it. The new lane is registered in `scripts/smoke.mjs` so it is part of the matrix rather than a file nobody runs.

Everything here is `#[ignore]` and credential-gated: without a throw-away bucket these skip and pass, and measure nothing.

## 2. Cloudflare R2: the jurisdictions were unreachable

An R2 bucket created in a jurisdiction answers only on its own host. The preset composed `{accountId}.r2.cloudflarestorage.com` with no way to say otherwise, and a search across the tree found no `.eu.r2` or `.us.r2` anywhere: **`eu` was missing too**, not only the `us` announced on 2026-08-17. The two ways to reach such a bucket today are typing the segment into the Account ID field or unlocking the suffix and writing the endpoint by hand, which is a workaround, not support.

The suffix beside the Account ID becomes a picker showing the host each choice produces. The code is stored bare (`eu`) and mapped to a host segment in one place per side, both **total**: absent or unknown means the default placement, never a failure to resolve. That is what keeps profiles saved before this option existed working, since an unresolved placeholder makes the resolvers return nothing at all.

It reaches the surfaces without a form, as a connection option has to: the profile loader (CLI, MCP pool, schedulers), the CLI mode scrub, and the bridge's endpoint-to-preset mapping, which keeps recognising the longer host as R2 instead of falling through to `custom-s3`.

**One defect was introduced and closed inside this change.** The form recovers an Account ID from an endpoint saved before that field existed by building a regex from the template, substituting the one placeholder. With a second placeholder present, that regex looked for the literal characters `{jurisdiction}` in the host and matched nothing, which would have silently stopped recovering the account id of every R2 profile in the old format. `parseS3EndpointParams` now builds the pattern by splitting the template on its placeholders, handles any number of them, and recovers the jurisdiction as well.

## Gate

`cargo fmt --all`, `cargo clippy --all-targets -- -D warnings` (the CI command verbatim) green, `cargo test` green, `npx tsc --noEmit` clean, `npm run i18n:validate` at 0 placeholders (no new i18n keys: provider preset labels are plain text in the registry, like every other field there), 10 new frontend tests, 3 new backend tests in the profile loader, 2 jurisdiction hosts added to the bridge mapping test.

A compile error in the new S3-compat test was hidden by a `grep` in a build pipeline that returned 0 regardless; it was caught by rebuilding without the filter, which is why the lane is verified to compile and to register both of its tests rather than assumed to.

## Summary by CodeRabbit

- **New Features**
  - Added Cloudflare R2 jurisdiction support, including jurisdiction-specific endpoint selection for EU and US regions.
  - Added jurisdiction selection to S3-compatible connection profiles.
  - Existing profiles now preserve and recover jurisdiction settings when edited.
  - Speed tests and connection setup use the selected jurisdiction automatically.
  - Added jurisdiction configuration support to the CLI.

- **Bug Fixes**
  - Improved recognition of jurisdiction-specific Cloudflare R2 endpoints.
  - Preserved endpoint behavior for custom and non-R2 S3 providers.

- **Tests**
  - Expanded Backblaze B2 testing to verify server-reported digests, large-file renames, and S3-compatible connections.

### DIFF
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -51,7 +51,7 @@ Legend:
 
 | Provider | Unit | Live (auto) | Live (manual) | Notes |
 |----------|:----:|:-----------:|---------------|-------|
-| Backblaze B2 | yes | yes (multipart 250 MB, 5.1 GB rename) | env-gated CI | Best automated cloud coverage |
+| Backblaze B2 | yes | yes (multipart 250 MB, 5.1 GB rename, server-reported digests) | env-gated CI | Best automated cloud coverage. Both code paths: the native B2 API and the S3-compatible endpoint (`integration_b2_s3_compat`, needs `AEROFTP_TEST_B2_S3_ENDPOINT` as well) |
 | Google Drive | yes | - | v4.0.0 (multipart 500 MiB) | |
 | Dropbox | yes | - | v4.0.0 (upload sessions) | |
 | OneDrive | yes | - | v4.0.0 (session upload) | |
--- a/scripts/smoke.mjs
+++ b/scripts/smoke.mjs
@@ -109,6 +109,14 @@ const integrationLanes = [
       why: 'set AEROFTP_TEST_B2_KEY_ID, AEROFTP_TEST_B2_KEY, AEROFTP_TEST_B2_BUCKET to enable',
     },
   },
+  {
+    name: 'integration_b2_s3_compat',
+    desc: 'b2 through its s3-compatible endpoint',
+    gate: {
+      probe: async () => hasEnv('AEROFTP_TEST_B2_KEY_ID', 'AEROFTP_TEST_B2_KEY', 'AEROFTP_TEST_B2_BUCKET', 'AEROFTP_TEST_B2_S3_ENDPOINT'),
+      why: 'set AEROFTP_TEST_B2_KEY_ID, AEROFTP_TEST_B2_KEY, AEROFTP_TEST_B2_BUCKET, AEROFTP_TEST_B2_S3_ENDPOINT to enable',
+    },
+  },
   { name: 'dag_s3_multipart', desc: 's3 multipart upload dag', gate: VAULT },
   { name: 'dag_s3_server_side_copy', desc: 's3 server-side copy dag', gate: VAULT },
   { name: 'dag_webdav_copy', desc: 'webdav copy dag', gate: VAULT },
--- a/src-tauri/src/bin/aeroftp_cli.rs
+++ b/src-tauri/src/bin/aeroftp_cli.rs
@@ -23368,6 +23368,7 @@ fn scrub_options_for_mode(opts: &mut serde_json::Map<String, serde_json::Value>)
         "anonymous",
         "bucket",
         "accountId",
+        "jurisdiction",
         // WebDAV mode
         "webdavScheme",
         "verifyCert",
--- a/src-tauri/src/bridge_shared.rs
+++ b/src-tauri/src/bridge_shared.rs
@@ -668,6 +668,17 @@ mod tests {
             map_s3_provider_from_endpoint("https://abc.r2.cloudflarestorage.com"),
             "cloudflare-r2"
         );
+        // A jurisdiction-scoped bucket has an extra segment in the host. It is
+        // still R2, and the bridge must not fall through to `custom-s3` and
+        // lose the preset's defaults (path-style, region `auto`).
+        assert_eq!(
+            map_s3_provider_from_endpoint("https://abc.eu.r2.cloudflarestorage.com"),
+            "cloudflare-r2"
+        );
+        assert_eq!(
+            map_s3_provider_from_endpoint("https://abc.us.r2.cloudflarestorage.com"),
+            "cloudflare-r2"
+        );
         assert_eq!(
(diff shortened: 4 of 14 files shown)


axpnet:
Thanks. One of the three is fixed, two are declined here with a reason.

**Cleanup before the assertion: fixed** in `ad0276ce8`, and it was a real defect of mine. The failure those assertions exist to catch is the one expected to happen one day, so a panic above the delete leaves the run's objects behind in exactly the case the test was written for, and in the multipart test that leftover is a 250 MB object on a free account with a 10 GB cap. All five now read the digest, clean up, then assert. Rerun against the throw-away bucket afterwards: same five digests as before the restructure.

**Credentials through a saved profile instead of the environment: not in this pull request.** The point is fair, but the sibling lane `integration_b2` has resolved the same three variables this way since it was written, and converting only the new file would leave two conventions for the same service, with the shared `AEROFTP_TEST_B2_*` triple split across them. Two things also bound what the change would buy: no workflow in this repository exports those variables, so the lane skips in CI and the keys are never in a job environment; and moving them to the vault puts them on the same machine under the same user, which changes where the secret sits and not who can read it. Worth doing as its own change covering both files, not as a side effect of this one.

**Rejecting `http://` custom S3 endpoints before signing: valid, and outside this diff.** `build_url` and `sign_request` live in `providers/s3.rs`, which this pull request does not touch; the test simply passes an endpoint through. The concern is real and it has a precedent here: Swift already carries an explicit cleartext consent (`allow_cleartext_storage_endpoint`), and S3 has no equivalent. A flat rejection would break the loopback and LAN bridges that legitimately speak HTTP (MinIO, Filen Desktop's local S3), so the right shape is probably that same explicit flag rather than a hard refusal, which makes it a design change with its own scope. Recorded for that.

PR #726, opened 2026-09-04
license(aerorsync): state the crate's licence, and give the emitted package the metadata it needs

axpnet:
## The gap

The emitted crate declared no licence at all. `[package]` carried no `license` field, no licence text was copied into the package, and of the 25 files in the module only one had an SPDX header, which said `GPL-3.0-or-later` while the reserved repository `axpdev-lab/aerorsync` is MPL-2.0. Nothing complained, because a crate with no licence header compiles exactly like a crate with one.

This closes correction 6 of the plan review (`provenance and licence gate before E1`), which was open by choice and had the measurable half already measured.

## The expression

`MPL-2.0 OR GPL-3.0-or-later`, stated in three places that cannot drift apart: an SPDX tag in every one of the 25 source files, `[package].license` in the emitted manifest, and the two verbatim licence texts shipped inside the package as `LICENSE-MPL-2.0` and `LICENSE-GPL-3.0`.

The same source has two consumers and each needs a different answer. MPL-2.0 is the licence of the reserved repository and the right fit for a protocol library other projects should be able to link. GPL-3.0-or-later is what the same files are as they compile into this application. The disjunction serves both and reopens neither decision. It also makes the MPL-2.0 section 3.3 question moot, because the GPL option is stated outright instead of being reached through the secondary-licence clause; Exhibit B (`Incompatible With Secondary Licenses`) is applied to no file.

## The gate

`scripts/aerorsync-emit.sh` now refuses to produce a package when a file is missing the tag, and names the file. A rule without a gate is a convention, and the failure it guards against arrives through a new file that nobody connects to this PR.

Verified by breaking it rather than by reading it: deleting the tag from `frame_io.rs` makes the emission exit 1 with

```
emissione rifiutata: file senza '// SPDX-License-Identifier: MPL-2.0 OR GPL-3.0-or-later': frame_io.rs
aggiungi l'header al sorgente in .../src/aerorsync, non alla copia emessa
```

pointing at the source rather than at the emitted copy, which is where the fix belongs.

## Provenance, and the limit of the clean-room claim

`LICENSING.md` ships with the crate. It records that the 25 files carry four e-mail identities in the git history, all belonging to the project owner, and that PR #97, the only known external contribution to this repository, never touched this module. Sole authorship is what makes the dual statement available to state.

It also bounds a claim that was drifting. The clean-room claim is about the **protocol**: implemented against captured bytes and published specifications, with no rsync source read into the tree, and the wire format is an interface rather than an expression. It is **not** about `scripts/aerorsync-emit.sh`, which performs a mechanical copy of these files into a standalone package. That is a build step, not an independent re-implementation, and calling it clean-room would stretch a true claim over something it does not describe. The module README now says so.

## Evidence

| Gate | Result |
|---|---|
| `cargo fmt --all -- --check` | rc=0 |
| `cargo clippy --all-targets -- -D warnings` | rc=0 |
| `cargo test --lib aerorsync` | 692 passed, 0 failed, 15 ignored |
| `bash scripts/aerorsync-emit.sh` | rc=0, 52 entries, package carries the licence field, both texts and 25 tagged files |
| Guard, tag removed from one file | rc=1, names the file, points at the source |

No behaviour changes: the diff is comment headers, two new files that ship with the package, and the emitter step that writes and checks them.

## Summary by CodeRabbit

- **Documentation**
  - Added licensing and provenance documentation for the AeroSync component.
  - Included the complete MPL-2.0 license text and clarified dual licensing under MPL-2.0 and GPL-3.0-or-later.
  - Added licensing notices across AeroSync source files.

- **Build and Distribution**
  - Generated packages now include the project README, licensing files, and additional package metadata.
  - Package generation validates required SPDX headers before creating output and stops with an error when they are missing.

---

## Update: two commits added after review of the first

**`fix`: the guard ran after the destructive step.** The licence check added above ran after `rm -rf "$OUT"` and after the module was copied, so a refusal left a half-written destination: sources but no manifest, no `lib.rs`, no licence texts, which resembles a crate closely enough to be mistaken for one. A gate placed after the destructive step protects the next step, not this one. It now runs before the wipe and reads the source tree, which is where the error message already pointed. Verified both ways: with the tag removed the run exits 1 and the destination has zero entries; with it restored the emission produces manifest sha256 `088ad700...`, the same value as before the move, so what is emitted is byte for byte what it was.

**`feat`: package metadata.** The manifest named the crate and its dependencies and nothing else.

| Field | Why it is not cosmetic |
|---|---|
| `rust-version = "1.85.0"` | Without it a consumer on an older toolchain gets a compile error inside the module instead of the sentence naming the required version. Read from the application manifest, not copied, for the same reason the dependency versions are read: a copy drifts on the next MSRV bump and the drift surfaces as an error nobody connects to this script |
| `readme = "README.md"` | The field names a file, so the module README now ships in the package. A field naming an absent file is worse than no field, because `cargo package` only rejects it at publication |
| `keywords`, `categories` | What the package looks like to whoever receives it |

**Deliberately not here**, because both are public surface and the surface is what D1 freezes: the `aerorsync::aerorsync::` double namespace that `lib.rs` produces, and the feature matrix, which the crate does not have at all. Fixing either now would decide by accident what D1 exists to decide, and D-D5 is open inside that same question. `publish = false` stays for the same reason: that flag flips in E1.

`cargo check --lib` inside the emitted crate: rc=0.

**Gate that was missing from the list above.** `scripts/aerorsync-emit-guard-tests.sh` covers the emitter's destination refusals, and this PR edits that emitter, yet the first evidence table did not name it: `fmt`, `clippy`, `cargo test` and a successful emission all pass without ever exercising a refusal scenario. Run here: **42 scenarios passed, 0 failed**, rc=0. The omission is the point worth recording, not the result: a change to a script is not covered by the gates that compile the code the script emits.


axpnet:
All three findings applied in `064a7d145`. None rejected.

**Provenance.** The finding is right and the wording was mine: the section claimed sole authorship and called it measured. What was measured is that every commit touching the module is authored under one of four e-mail addresses, all the owner's, and that a spot check of PR #97 shows it changed a file outside this module. That is evidence about who committed, not a proof that no third-party code was incorporated, and one pull request is not an enumeration of every external contribution. Dressing an inference as a measurement made the weaker claim read as the stronger one.

The section now separates three things that were run together: what the history records, what that does and does not establish, and what actually licenses the work. On the last point the finding is also right that the repository uses DCO rather than a CLA, so the basis is now stated as it is: each contribution carries a `Signed-off-by` in which its author certifies the right to submit it under the project's licence, and that together with the copyright the owner holds in their own contributions is what makes the dual expression theirs to state.

**SPDX scope.** Also correct, and the same shape of error. The guard checks `*.rs`; the module also tracks capture scripts, Dockerfiles, compose files, frozen transcript artifacts and these documents, none of which carry a Rust comment header. The claim is now bounded to Rust sources, the untagged files are named, and the reason they are covered by the package-level licence instead is stated. The identical sentence in the module README is narrowed too, because a claim repeated in two places has to be corrected in both.

**MD040.** Fence given a language.

Re-verified after the change: emission rc=0, and the 42 scenarios in `scripts/aerorsync-emit-guard-tests.sh` pass.

PR #733, opened 2026-09-04
fix(s3): refuse to sign for a cleartext endpoint without consent

axpnet:
## The gap

The cleartext consent gate added in #729 lives in `connect()`. `discover_buckets()` does not go through `connect()`: it is `pub`, Quick Connect reaches it directly from `provider_discover_targets` (`provider_commands.rs:1414`), and it ran `ensure_fresh_credentials()` and then `sign_request()` against whatever endpoint the user had just typed.

So on a plain `http://` endpoint that is not on this machine, pressing the button that lists buckets sent a signed request with the `Authorization` header readable on the wire, and the consent the profile is supposed to require was never asked for. That is precisely the exposure #729 was written to close, reached through a door it did not cover.

Verified against `origin/main` before anything was written:

```
validate_endpoint_confidentiality   1 production call site (s3.rs, inside connect)
                                    8 call sites in tests
provider_commands.rs:1411-1415      S3Provider::new(cfg).discover_buckets()   <- no connect()
discover_buckets                    ensure_fresh_credentials -> sign_request("GET") -> send
```

## The fix is at the choke point, not at the door

The obvious repair is a call at the top of `discover_buckets`. This puts it in `sign_request()` instead, and that difference is the point: guarding one door at a time is exactly how this door came to be missed.

A signature authenticates a request without encrypting it, so the `Authorization` header `sign_request` is about to build is itself the secret at risk. Ten of the eleven signing paths in `s3.rs` pass through it, and a twelfth written six months from now inherits the check without anyone having to remember it exists. The eleventh is the query signer described below, which needed its own call.

`connect()` keeps its own call, so a session fails before any work rather than at its first request.

There are two other places the check has to be, and review found both. **The STS exchange runs before anything is signed for S3**, so on a door that does not go through `connect()` a role profile burnt an AssumeRole, and a one-time MFA code with it, then failed with an STS error instead of the reason it was refused. Not a cleartext exposure, since STS always dials `sts.<region>.amazonaws.com` over https, but the wrong work and the wrong message. The guard is now in `ensure_fresh_credentials` too, after the role and freshness early returns, so a profile without a role pays nothing for it. The first version of this body claimed that refusing in `connect()` is what stops the STS exchange: true for `connect()`, false for this door.

**And there is an eleventh signer that does not go through `sign_request` at all**: `create_share_link` builds a SigV4 query signature rather than an Authorization header. A presigned URL for a cleartext endpoint is worse than a single request, because it is a bearer credential that travels readable every time anyone opens it. Guarded at the top of that method.

## The sibling doors were checked, not assumed

Closing one door and leaving its twin would be the same defect applied to its own remedy, so `provider_discover_targets` was read in full. It has three branches:

| Branch | Exposure | Why |
|---|---|---|
| S3 | yes, fixed here | `S3Config` carries a user-supplied `endpoint` |
| Backblaze | none | `B2Config` has four fields and no endpoint at all, so discovery talks to Backblaze's fixed HTTPS API |
| kDrive | none | pinned to `https://api.infomaniak.com` by a constant |

Swift has a cleartext guard of its own (`allow_cleartext_storage_endpoint`) and no discovery branch here, so it is untouched.

## The tests are on the door, not on the guard

Eight tests already assert that `validate_endpoint_confidentiality` refuses a cleartext endpoint. All eight passed while discovery signed and sent anyway, so a ninth of the same kind would have proved nothing. The difference between testing the guard and testing the door is this entire finding.

- `bucket_discovery_sends_nothing_to_a_cleartext_endpoint_without_consent` asserts that no request leaves. The endpoint host is in `.invalid`, which is reserved and never resolves, so the two outcomes are distinguishable rather than merely worded differently: with the guard the error is the confidentiality refusal, without it the DNS failure surfaces as a `NetworkError` naming ListBuckets, which is the request already in flight.
- `bucket_discovery_proceeds_once_the_profile_consents` is the control. Without it the first test would also pass on a `discover_buckets` that refused everything unconditionally.
- `signing_refuses_a_cleartext_endpoint_without_consent` pins the choke point itself and asserts that no `Authorization` header is handed back to a caller that was refused.
- `a_role_profile_is_refused_before_the_sts_exchange` asserts a role profile gets the confidentiality refusal and not an `AssumeRole` failure, which is what proves the exchange did not happen.
- `a_share_link_is_refused_for_a_cleartext_endpoint_without_consent` covers the query signer.
- `the_cleartext_warning_is_logged_once_per_provider` counts WARN events through a tracing layer. The consent branch logs, and moving the check into `sign_request` turned one line per session into one per signature. **An earlier draft of this test asserted the once-slot instead, which is the guard and not the door**: removing the condition around the warning would have restored one line per signature and left it green, which is the same mistake this pull request exists to fix, made half an hour after writing it down.

**Fail-first, and it is worth quoting.** Run against the unfixed code, the door test fails with:

```
the refusal must be the confidentiality one, got
NetworkError("ListBuckets failed: error sending request for url (http://s3.invalid:9000/)")
```

That message is the vulnerability demonstrated rather than argued: the signed request was already on the wire. The control test passed in that same run, as designed.

## Gate

The Tier 1 pre-push gate in full: `cargo fmt --all --check`, `npm run typecheck`, `npm run i18n:validate`, `npm run i18n:diacritics`, `npm run i18n:untranslated`, `npm run check:provider-inventory`, `npm run test:unit`. Plus `cargo clippy --all-targets -- -D warnings` and `cargo test --lib s3` with 156 tests.

Fail-first covers every remedy and not only the first: four mutations, one per guard, each caught by its named test. One first reported as NOT CAUGHT, and that was the harness rather than the code, since the filter it ran did not match the test's name. A mutation runner that cannot tell "the test did not fail" from "the test did not run" is the same blind guard the rest of this pull request is about.

## Provenance

Reported by CodeRabbit as an out-of-diff comment while reviewing #732, which is an unrelated change to the delta grid. It was verified against `origin/main` line by line before being acted on, and deliberately kept out of #732: a security fix folded into a pull request about part sizing would make both stories unreadable, and would make this one invisible to anyone later asking when it was closed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Cleartext storage endpoints are now blocked consistently during request signing unless explicit profile consent is provided.
  - Bucket discovery validates endpoint security before making network requests.
  - Presigned share links now apply the same cleartext endpoint protection.
  - Cleartext warnings are shown only once per provider connection, reducing duplicate alerts.
  - Requests to approved endpoints continue to work normally.

### DIFF
--- a/src-tauri/src/providers/s3.rs
+++ b/src-tauri/src/providers/s3.rs
@@ -325,6 +325,12 @@ pub struct S3Provider {
     /// Serializes STS refreshes so concurrent multipart parts re-assume the
     /// role once instead of stampeding the STS endpoint near expiry.
     sts_refresh_lock: Arc<tokio::sync::Mutex<()>>,
+    /// One cleartext warning per provider, not one per signed request.
+    /// The confidentiality check runs at every signing path now, and its
+    /// consent branch used to log unconditionally, so an N part upload logged
+    /// N identical warnings. Shared across clones on purpose: a clone is the
+    /// same profile talking to the same endpoint.
+    cleartext_warned: Arc<std::sync::atomic::AtomicBool>,
 }
 
 /// Owned snapshot of the credentials a single signing pass uses, taken under
@@ -403,6 +409,7 @@ impl S3Provider {
             storage_class_override: None,
             temp_credentials: Arc::new(std::sync::RwLock::new(None)),
             sts_refresh_lock: Arc::new(tokio::sync::Mutex::new(())),
+            cleartext_warned: Arc::new(std::sync::atomic::AtomicBool::new(false)),
         })
     }
 
@@ -563,6 +570,16 @@ impl S3Provider {
         if !self.temp_credentials_need_refresh() {
             return Ok(());
         }
+
+        // Second choke point, and it is not the same as the one in
+        // `sign_request`. An AssumeRole is a signed request of its own, made
+        // before anything is signed for S3, so on a door that does not go
+        // through `connect()` a role profile would burn an exchange (and a
+        // one-time MFA code) against an endpoint the user never consented to,
+        // then fail with an STS error instead of the confidentiality one.
+        // Placed after the two early returns above, so a profile without a
+        // role or with fresh credentials pays nothing for it.
+        self.validate_endpoint_confidentiality()?;
 
         let _guard = self.sts_refresh_lock.lock().await;
         // Another task may have refreshed while we waited for the lock.
@@ -777,18 +794,34 @@ impl S3Provider {
             )));
         }
 
-        if self.config.role_arn.is_some() {
-            warn!(
-                "[S3] endpoint {} is CLEARTEXT HTTP and carries an STS session token:                  token and payload are unencrypted (opted in on this profile)",
-                host
-            );
-        } else {
-            warn!(
-                "[S3] endpoint {} is CLEARTEXT HTTP: credentials and payload are                  unencrypted (opted in on this profile)",
-                host
-            );
+        // Once per provider. This runs at every signing path, so an
+        // unconditional log here would put one identical warning per part of a
+        // multipart upload into the user's log and drown the rest of it.
+        if self.take_cleartext_warning_slot() {
+            if self.config.role_arn.is_some() {
+                warn!(
+                    "[S3] endpoint {} is CLEARTEXT HTTP and carries an STS session token:                  token and payload are unencrypted (opted in on this profile)",
+                    host
+                );
+            } else {
+                warn!(
+                    "[S3] endpoint {} is CLEARTEXT HTTP: credentials and payload are                  unencrypted (opted in on this profile)",
+                    host
+                );
+            }
         }
         Ok(())
+    }
+
+    /// True the first time it is called on a provider and false afterwards.
+    ///
+    /// Split out so the once-ness is testable without a tracing subscriber:
+    /// asserting on the slot is asserting on exactly the decision the log line
+    /// depends on.
+    fn take_cleartext_warning_slot(&self) -> bool {
+        !self
+            .cleartext_warned
+            .swap(true, std::sync::atomic::Ordering::Relaxed)
     }
 
     /// Detect MEGA S4 Object Storage endpoints.
@@ -3846,6 +3879,14 @@ impl StorageProvider for S3Provider {
         path: &str,
         options: ShareLinkOptions,
     ) -> Result<ShareLinkResult, ProviderError> {
+        // The eleventh signer, and the only one that does not go through
+        // `sign_request`: this builds a SigV4 QUERY signature instead of an
+        // Authorization header, so it inherits nothing from the choke point
+        // below. A presigned URL for a cleartext endpoint is worse than a
+        // single request, because it is a bearer credential that travels in
+        // the clear every time anyone opens it.
+        self.validate_endpoint_confidentiality()?;
+
         // Generate a presigned URL
         use hmac::{Hmac, Mac};
         use sha2::{Digest, Sha256};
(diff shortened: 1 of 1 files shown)


axpnet (commented):
Adversarial review of `a8d840794` (Claude L2, review tab). Looked at: the diff, `validate_endpoint_confidentiality` (s3.rs:750), `sign_request` (s3.rs:941), `discover_buckets` (s3.rs:1243), `ensure_fresh_credentials` (s3.rs:559), `connect` (s3.rs:2712), `create_share_link` (s3.rs:3844), `send_with_retry` (http_retry.rs:91), `sts.rs:184`, and every non-test constructor of `S3Provider` outside s3.rs. Ran `cargo test --lib s3` on the head: 153 passed, matching the body. The two door tests are fast because a transport error is returned by `send_with_retry` before the retry loop (http_retry.rs:113 to 120), so the `.invalid` host does not pay the backoff. Not looked at: the frontend, the Swift twin.

The fix is right and the door test is the right kind of test. Three things the body does not sustain, two of them measured with throwaway tests on this head (not committed), listed by weight.

**1. Moderate, behaviour regression for consenting profiles: the warning now fires on every signature.** `validate_endpoint_confidentiality` ends with an unconditional `warn!` when consent is present (s3.rs:780 to 789), and this PR calls it at the top of `sign_request`. Before, the line was logged once per connect. Now it is logged once per signed request, so a multipart upload of N parts logs it N times, plus once per list, head, delete. Measured: 5 calls to `sign_request` on a consenting provider produced 5 WARN events under a counting subscriber. Remedy: split the function into the silent predicate that refuses, used at the choke point, and the one-time warning kept in `connect()` (or behind a `std::sync::Once` on the provider). A test that signs twice on a consenting profile and asserts one warning would pin it.

**2. Minor, reachable: on the door this PR closes, a role profile still does the STS exchange before the refusal.** `discover_buckets` calls `ensure_fresh_credentials` (s3.rs:1253) before `sign_request` (s3.rs:1257), so with `role_arn` set the AssumeRole call goes out first and the guard runs after. Measured: `cleartext_test_provider("http://s3.invalid:9000", false, Some("arn:aws:iam::123456789012:role/review")).discover_buckets()` returned `AuthenticationFailed("STS AssumeRole denied: InvalidClientTokenId ...")`, which means the base key travelled to `sts.us-east-1.amazonaws.com` and the confidentiality message, the one that names the option to set, never reached the user. No cleartext exposure, STS is https (sts.rs:184), but it is a real AssumeRole per Quick Connect click on a profile that will be refused, an MFA TOTP code consumed for nothing (and `ensure_fresh_credentials` then refuses to replay it), and the wrong error on screen. The body sentence "refusing there is what stops the STS exchange" is true for `connect()` and false for the very door the PR is about. Remedy: the same guard in `ensure_fresh_credentials`, after the `role_arn` and freshness early returns and before the exchange. That is a second choke point rather than a door, it is only reached when a refresh is due so the hot path pays nothing, and the probe above becomes the test: with a role and no consent, `discover_buckets` must return the confidentiality refusal and no network error.

**3. Low, body accuracy: there is already an eleventh signer, and it does not inherit the guard.** "All ten signing paths in s3.rs pass through it" counts the callers of `sign_request`. `create_share_link` builds its own SigV4 query signature (s3.rs:3844 onward, `X-Amz-Signature` at 3962) and never calls `sign_request`. Today it is reachable only through a connected session (session_manager.rs:393, provider_commands.rs:6134, the CLI after connect), so the `connect()` guard covers it, but the argument of the PR is "guard where the signature is built", and this is a place where a signature is built without the guard. A presigned URL for a cleartext endpoint is a bearer credential that travels in the clear the moment it is used. Remedy: one call at the top of `create_share_link`, which has no network I/O of its own, and the body reworded to ten header signers plus one query signer.

Verified and holding: `provider_discover_targets` has the three branches the table says and only S3 carries a user endpoint; `provider_transfer_executor.rs:2376` constructs an S3Provider on loopback only inside a test; `endpoint_needs_cleartext_consent` is the single predicate on both sides (profile_export.rs); the `Authorization` header is not handed back on refusal because the check runs before any header is inserted.

PR #739, opened 2026-09-07
build(deps): move off the yanked der 0.8.0 and libssh2-sys 0.3.2

axpnet:
The main branch's Checks run went red on cargo audit (two denied warnings: der 0.8.0 and libssh2-sys 0.3.2 were yanked after the last lock update; the step is advisory on pull requests, authoritative on main). This bumps only those two entries in Cargo.lock to the latest compatible versions (der 0.8.2, libssh2-sys 0.3.3). cargo audit clean locally, tree type-checks; GitHub Actions decides.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -2272,9 +2272,9 @@ dependencies = [
 
 [[package]]
 name = "der"
-version = "0.8.0"
+version = "0.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "71fd89660b2dc699704064e59e9dba0147b903e85319429e131620d022be411b"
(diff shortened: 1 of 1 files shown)


axpnet:
Noting a dependency, since it is not visible from this pull request: this one currently gates the whole branch queue.

`main` is red at `bffa68256`: the *Checks* run fails on `cargo audit` with exactly these two yanked crates ([run 34105024772](https://github.com/axpdev-lab/aeroftp/actions/runs/34105024772)). It is not that commit's code, it is an upstream yank, so every run from now on fails the same way, on `main` and on any branch that re-runs after it.

I reached the same two updates independently from #738 and stopped when I found this pull request already open with them, so nothing is duplicated here. #738 is green but its green predates the yank and was measured against `main` at `7b3817c87`. It is held until this merges, then rebased on top so its next run is a real one.

No action needed beyond merging when it goes green.

PR #766, opened 2026-09-09
feat(s3): remember the bytes we uploaded, and match a delta against them

axpnet:
Steps T4 and T5 of the S3 delta upload lane, after the pure planner (#723) and the multipart executor (#755).

## What the lane gains

A delta upload needs to know what the remote object already contains. Reading the baseline back would cost egress on every attempt, so this keeps a local memory of the bytes this client uploaded: a SQLite store of per-cell digests, capped at 10,000 rows with LRU eviction and a per-row digest cap, keyed by the corrected endpoint and account identity plus bucket and object. Ordinary streaming uploads, the single PUT path and the DAG trait multipart path all seed it, and only after the upload response ETag certifies what landed. A failed, aborted or ETag-less upload leaves no row.

The matcher then aligns local cells against that memory without a single remote read, and the adapter wires S3 into the existing delta entry point. S3 failures are classified from the HTTP status and the XML error code, never from prose, and never through the native rsync string classifier: an authentication or permission refusal stays hard, everything else falls back to a plain upload, so a `503 SlowDown` cannot kill a transfer that has the ordinary path beside it. A backend that refuses ranged copy is remembered for 300 seconds, capped at 128 endpoints, and only when a Copy operation is the one that refused.

## What I added on top of the implementation commit

I inherited this branch with the code written but never compiled, so the second commit is what running it for the first time produced, plus the doors a mutation battery cannot reach on its own.

`MultipartBaseline::finish` writes a row only when every part begins on a grid boundary, because the DAG uploads parts on independent clones and hashes each one on its own, so the digests only concatenate where the cuts line up. That holds for the defaults, which is why it held wherever anyone looked, and it did not hold when the CLI's upload buffer override set a part size the grid cannot divide: no DAG upload seeded the store at all, the next delta answered `no_baseline`, and nothing said why. Failing safe and failing silently at once is the shape that never gets found, so the part size handed to the DAG is now snapped up to a whole number of grid cells, at a cost of at most one cell. Only the hint is snapped: the streaming path carries a partial cell across reads and is indifferent to part boundaries, which `s3_baseline_hasher_ignores_upload_chunk_boundaries` already pins, so the override is still honoured verbatim where it always worked. Above roughly 80 GB the grid grows past `DELTA_PART_SIZE` and the DAG builder grows the chunk itself to keep the part count legal, so alignment there is a declared limit rather than a promise. The domain table pins the precondition regardless, along with part 0, a number past the last legal one, a completion over the digest cap, non-consecutive numbers, an ETag the server completed under that is not the one we sent, incomplete coverage and an empty part, with the aligned control asserted first so a `finish` that always returned `None` cannot pass.

Nothing changed the source between the plan and the completion, so removing the pre-Complete guard stayed green by construction. The mock now grows the local file while the first plain PUT is in flight, and the door pins no completion, an abort, the object left byte for byte as it was, a soft fallback rather than a hard error, and no baseline row certifying bytes that completed nowhere.

The negative ranged-copy memory only proved that a Copy refusal is remembered, never that anything else is not, so moving the operation match to any other operation left the whole battery green. The converse now pins that a HeadObject, CreateMultipartUpload, UploadPart or CompleteMultipartUpload answering 501 or NotImplemented teaches nothing about ranged copy.

Two smaller things. The adapter reported a planner refusal as `no_match_found`, the same reason it uses when the matcher found nothing, which would have made the wire-ratio report count the two together; it is now `plan_refused`. And `RsyncStats::speedup` is documented as a wire ratio with a one byte floor on the denominator, not a wall-clock speedup.

Running the battery for the first time also surfaced two defects in the branch itself: the in-loop append of the wire-ratio door was never flushed or dropped, so the file was still growing when the delta planned over it and the source guard correctly answered `source_changed`, and clippy refused the baseline file for items after a test module.

## Mutation battery

Four mutations, four caught, none unmeasured. Each was applied to production code only, scoped to the function under test, and reverted afterwards.

| # | Mutation | Test that went red |
|---|---|---|
| 1 | Neutralize the pre-Complete source guard in the typed executor | `s3_adapter_refuses_to_complete_when_the_source_changes_mid_upload` |
| 2 | Match `DeltaOperation::Head` instead of `Copy` in both arms of `remember_range_rejection` | `s3_adapter_negative_memory_only_remembers_a_copy_refusal` |
| 3 | Drop the grid-boundary check from `MultipartBaseline::finish` | `s3_baseline_multipart_finish_rejects_every_completion_it_cannot_certify` |
| 4 | Drop the part-number domain from `MultipartBaseline::record` | `s3_baseline_multipart_record_rejects_part_numbers_outside_the_s3_domain` |

Mutation 2 is worth a note on method: `operation: DeltaOperation::Copy` appears five times in that file, not twice, because the tests contain it too. Applying it file-wide would have mutated test code and returned a meaningless green. Scoping it to the function body is what makes the result mean anything.

## Declared gaps

The 128-entry cap of the negative memory has no test, deliberately. The map is a process-wide static shared by every test in the binary and tests run in parallel, so proving the cap means inserting 129 entries and evicting other tests' rows while they run. That trades a missing test for a flaky suite. It becomes testable if the cache is injectable rather than static, or if the test is serialized.

The write-side digest cap is unreachable from either hasher by construction, so it is two guards with no door; testing it needs a baseline built by hand at the boundary and one past it. The adapter's `file_too_large` refusal is not reachable through the door either: it would need a 25 TiB sparse file, past what ext4 addresses, so it is testable only if that triage becomes a pure function of the size.

The baseline row is invalidated before the executor runs, so a planner refusal or a transient error leaves no row and the fallback upload reseeds it. That is fail safe and the door asserts the outcome that matters (no row may certify bytes that never completed) rather than assuming the row survives.

No live S3 account was exercised for this branch. The doors run against a local HTTP mock, and the live lane stays for the step that turns the executor on by default.

## Verification

`s3_baseline_` 14 passed, `s3_adapter_` 8 passed, `s3_delta_plan` 41 passed, plus #764's own `recursive_delete_preserves` and `batch_delete_fallback` tests re-run after the rebase since both commits touch `s3.rs`. `cargo clippy --all-targets -- -D warnings` and `cargo fmt --all --check` clean, on the rebased tree.

## Summary by CodeRabbit

* **New Features**
  * Added S3 delta uploads that reuse unchanged portions of previously uploaded files to reduce data transfer.
  * Added local, account- and endpoint-aware baselines for faster repeat and appended-file uploads.
  * Added progress reporting and transfer statistics for S3 delta uploads.

* **Bug Fixes**
  * Improved fallback handling when delta uploads are unsupported or unavailable.
  * Added safeguards to detect files changing during transfer and prevent incomplete uploads.
  * Ensured interrupted or cancelled multipart uploads are properly aborted.
  * Improved validation of upload data and completion results.

## Cancellation safety (T5.1)

Cancelling a transfer from the GUI drops the transfer future: `provider_commands.rs` races it against the cancel token and the cancel arm wins by dropping it. Dropping a future runs no error branch at all, so every explicit `AbortMultipartUpload` in the executor was unreachable from the one path users actually take, and the parts already uploaded stayed under the upload id until the bucket lifecycle rule removed them, if the bucket had one. The object stayed atomic and intact; what leaked was remote storage, which S3 bills for.

`MultipartAbortGuard` owns the upload id from the moment the server returns it until a completion is proven. It has three entries, and the split is deliberate. Known failures call `abort_now` and await it, so the abort has happened before the caller sees the error and a test can assert the count. A proven completion calls `disarm`. `Drop` covers what no branch can reach, cancellation and panic, and since `Drop` is sync it schedules the abort on the current runtime, the shape `ProviderGuard` already uses for disconnect. The alternative, a single always-spawned owner, was measured against the tree and rejected: it would have made `s3_baseline_failed_aborted_and_missing_etag_uploads_leave_no_row` and `delta_412_from_the_guard_aborts_and_never_completes` racy, and it would have extended the guard's own warn-and-leak fallback, which fires when no runtime is left, to paths that do not need it.

The guard is armed on both multipart paths, the delta executor and `upload_multipart_streaming`. On the classic path it covers cancellation and nothing else: that function's explicit branches already aborted correctly, so the pre-existing exposure there was the drop case alone, not a forgotten branch.

Door test: the mock parks the first plain PUT, the test drops the transfer future while that part is genuinely in flight, then pins an abort, no completion, the object byte for byte as it was, and no baseline row. The abort is scheduled by `Drop` rather than awaited, so the test polls for it within an explicit five second bound and says so when it fails.

### Deferred, with the reason

Dropping the future *during* `CompleteMultipartUpload` is left as it is, and the outcome is bounded rather than unknown. The spawned DELETE races the in-flight completion and both orders are consistent: if the DELETE lands first the completion fails with NoSuchUpload, the parts are removed and the baseline object is untouched; if it lands second the completion won, the object is replaced and the DELETE gets a 404. Never a torn state, never an orphan part, never lost data. What is lost in the second order is the baseline row, which was already invalidated so the next delta simply uploads in full, and the accuracy of the GUI message, which says cancelled while the object is committed. That is accounting and cosmetics, not safety, and it does not justify widening this change. It is recorded as a T6 item with its own door test (drop during the completion: the object is committed and the row is written), alongside draining pending aborts on `disconnect`, which is the production justification that would also make the door test above stop polling.

## Mutation battery

Twelve mutations applied to production code on this exact tree, each scoped to the function under test, each reverted and the revert verified against a byte-for-byte backup rather than against HEAD, which would have been the wrong instrument while this work was uncommitted.

| # | Mutation | Test that went red |
|---|---|---|
| M1 | Route the S3 arm through the native rsync classifier | `s3_adapter_http_failures_fallback_except_auth_and_never_poison_412_capability` |
| M2 | Demote auth: drop `401 \| 403` and the credential code list together | `s3_adapter_classifier_uses_status_and_xml_code_not_prose` |
| M3 | Lose wire accounting on PUT parts | `s3_adapter_middle_append_and_successive_delta_have_independent_wire_ratios` |
| M3b | Count ranged copies as wire bytes | same |
| M4 | Do not certify the row with the completion ETag | same (`no_baseline` at door 1) |
| M4b | Seed the row from the pre-delta HEAD ETag | same (`etag_mismatch`) |
| M4c | Completion returns no ETag | `s3_baseline_upload_doors_certify_only_successful_response_etags` |
| M5 | Drop `x-amz-copy-source-if-match` from ranged copies | wire-ratio door (`etag_mismatch`) |
| M6 | Cache 412 and 429 in the negative capability memory | `s3_adapter_negative_memory_does_not_cache_preconditions_or_transient_errors` |
| MG | Neutralise the abort guard's `Drop` | `s3_adapter_dropping_a_delta_upload_aborts_the_multipart` |
| MH | Remove the snap that aligns the DAG part size hint | `dag_part_size_hint_is_always_a_whole_number_of_grid_cells` |
| MS | Neutralise the pre-completion source guard | `s3_adapter_refuses_to_complete_when_the_source_changes_mid_upload` |

Measured 12, caught 12, not measured 1. The one not measured is "pass a full upload off as a delta": the planner absorbs it, since empty matches return `None` and the attempt is refused before any request, so there is no mutation of the executor that reaches it. It is reported as not measured with that reason rather than forced into a shape that would prove nothing.

M2 carries a method note worth keeping: the status edit alone leaves the door green, because the mock's 403 also carries an `AccessDenied` code, so both halves belong in one edit. An earlier attempt at M6's sibling was stopped by an assertion when the string to mutate turned out to appear five times in the file rather than twice, the extra three being in tests: mutating test code would have returned a meaningless green.

### DIFF
--- a/.github/workflows/aerorsync-protocol.yml
+++ b/.github/workflows/aerorsync-protocol.yml
@@ -77,7 +77,7 @@ jobs:
         run: |
           # Runner-image extras: Microsoft's apt repos 403 on their InRelease often
           # enough to redden a run, and nothing here installs from them.
-          sudo rm -f /etc/apt/sources.list.d/microsoft* /etc/apt/sources.list.d/azure-cli*
+          sudo rm -f /etc/apt/sources.list.d/microsoft* /etc/apt/sources.list.d/azure-cli* /etc/apt/sources.list.d/google-chrome*
           sudo apt-get update
           sudo apt-get install -y --no-install-recommends \
             libwebkit2gtk-4.1-dev libgtk-3-dev libayatana-appindicator3-dev \
@@ -138,7 +138,7 @@ jobs:
         run: |
           # Runner-image extras: Microsoft's apt repos 403 on their InRelease often
           # enough to redden a run, and nothing here installs from them.
-          sudo rm -f /etc/apt/sources.list.d/microsoft* /etc/apt/sources.list.d/azure-cli*
+          sudo rm -f /etc/apt/sources.list.d/microsoft* /etc/apt/sources.list.d/azure-cli* /etc/apt/sources.list.d/google-chrome*
           sudo apt-get update
           sudo apt-get install -y --no-install-recommends \
             libwebkit2gtk-4.1-dev libgtk-3-dev libayatana-appindicator3-dev \
@@ -199,7 +199,7 @@ jobs:
         run: |
           # Runner-image extras: Microsoft's apt repos 403 on their InRelease often
           # enough to redden a run, and nothing here installs from them.
-          sudo rm -f /etc/apt/sources.list.d/microsoft* /etc/apt/sources.list.d/azure-cli*
+          sudo rm -f /etc/apt/sources.list.d/microsoft* /etc/apt/sources.list.d/azure-cli* /etc/apt/sources.list.d/google-chrome*
           sudo apt-get update
           sudo apt-get install -y --no-install-recommends \
             libwebkit2gtk-4.1-dev libgtk-3-dev libayatana-appindicator3-dev \
--- a/.github/workflows/aerorsync-standalone.yml
+++ b/.github/workflows/aerorsync-standalone.yml
@@ -73,7 +73,7 @@ jobs:
         run: |
           # Runner-image extras: Microsoft's apt repos 403 on their InRelease often
           # enough to redden a run, and nothing here installs from them.
-          sudo rm -f /etc/apt/sources.list.d/microsoft* /etc/apt/sources.list.d/azure-cli*
+          sudo rm -f /etc/apt/sources.list.d/microsoft* /etc/apt/sources.list.d/azure-cli* /etc/apt/sources.list.d/google-chrome*
           sudo apt-get update
           # `libacl1-dev` is not optional here: `acl-sys` declares
           # `#[link(name = "acl")]` on Linux, so the test binary does not link
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,7 +64,7 @@ jobs:
         run: |
           # Runner-image extras: Microsoft's apt repos 403 on their InRelease often
           # enough to redden a run, and nothing here installs from them.
-          sudo rm -f /etc/apt/sources.list.d/microsoft* /etc/apt/sources.list.d/azure-cli*
+          sudo rm -f /etc/apt/sources.list.d/microsoft* /etc/apt/sources.list.d/azure-cli* /etc/apt/sources.list.d/google-chrome*
           sudo apt-get update
           # `dbus` is here for `cargo test` / portal_chooser (#510), not for
           # packaging. `portal_chooser`'s three cases stand up a throwaway
@@ -238,7 +238,7 @@ jobs:
         run: |
           # Runner-image extras: Microsoft's apt repos 403 on their InRelease often
           # enough to redden a run, and nothing here installs from them.
-          sudo rm -f /etc/apt/sources.list.d/microsoft* /etc/apt/sources.list.d/azure-cli*
+          sudo rm -f /etc/apt/sources.list.d/microsoft* /etc/apt/sources.list.d/azure-cli* /etc/apt/sources.list.d/google-chrome*
           sudo apt-get update
           # `dbus` lives on `linux-validate` with `cargo test` / portal_chooser
           # (#510). This leg only needs the Tauri/bundle toolchain.
@@ -832,7 +832,7 @@ jobs:
         run: |
           # Runner-image extras: Microsoft's apt repos 403 on their InRelease often
           # enough to redden a run, and nothing here installs from them.
-          sudo rm -f /etc/apt/sources.list.d/microsoft* /etc/apt/sources.list.d/azure-cli*
+          sudo rm -f /etc/apt/sources.list.d/microsoft* /etc/apt/sources.list.d/azure-cli* /etc/apt/sources.list.d/google-chrome*
           sudo apt-get update
           sudo apt-get install -y --no-install-recommends squashfs-tools binutils file
           ./scripts/snap-abi-check.sh "$SNAP_FILE"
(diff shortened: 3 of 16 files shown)


axpnet:
@coderabbitai full review

PR #773, opened 2026-09-10
chore(deps): take the semver-compatible updates and state the real MSRV

axpnet:
Takes every semver-compatible dependency update in one lockfile batch, keeps the Tauri JS and Rust packages in lockstep, and makes `rust-version` state the toolchain the resolved graph actually needs. No semver-incompatible bump is in here: those go in separate pull requests, one cohort each.

## Cargo

`cargo update`: 207 packages updated, 36 added, 21 removed in `Cargo.lock`. The ones worth naming:

- `russh` 0.63.1 to 0.63.3, `rustls` 0.23.44, `rustls-webpki` 0.103.15, `reqwest` 0.13.5, `hyper` 1.11.1, `aws-lc-rs` 1.18.1 with `aws-lc-sys` 0.45.0
- `iroh` 1.0.3 to 1.2.0 (peer-l0), `sevenz-rust2` 0.21.6, `toml` 1.1.5, `uuid` 1.26.1, `futures` 0.3.34
- `tauri-runtime-wry` 2.11.3 to 2.11.4 under the unchanged `tauri =2.11.0` pin. Its only source change is in the Windows DPI helper, which now releases the device context it acquires (`GetDC` paired with `ReleaseDC`); the crates were diffed, nothing else moved.
- `tauri-plugin-notification` 2.3.3 to 2.4.0, with the manifest floor raised to match.

## npm

Lockfile generated with npm 10.8.2, the version CI installs with.

- `@tauri-apps/plugin-notification` 2.4.0, in the same change as the crate: Tauri refuses to build when a plugin crate and its JS package differ in minor version.
- `dompurify` 3.4.15, `lucide-react` 1.43.0, `react-virtuoso` 4.18.13, `react-window` 2.3.1, `@types/react` 18.3.31, `@tauri-apps/plugin-fs` 2.5.2.
- `@tauri-apps/api` and `@tauri-apps/cli` are deliberately left where they are, in lockstep with the `tauri` pin.

## rust-version

`src-tauri/Cargo.toml` said 1.85.0, and that was already false on `main` before this change: 65 locked crates declared a higher `rust-version`, the highest being `sevenz-rust2` at 1.93, so no 1.85 toolchain could build the tree. It now says 1.93, measured as the highest `rust-version` any locked crate declares in `cargo metadata`, and `peer-l0` says 1.91 (the iroh stack). The toolchain that builds and lints stays pinned in `rust-toolchain.toml` (1.97.0).

This crate is on resolver 2, which does not consider `rust-version` when resolving, so a future compatible update can raise the floor again without a warning. The value has to be re-measured after every lockfile batch.

The dependabot hold on `globset >=0.4.20` is removed: its only stated reason was the 1.85.0 MSRV (globset 0.4.20 declares 1.88).

## A red this PR caused, and why the local gate did not see it

The first head (`db4acbe3a`) failed `cargo clippy -D warnings` in CI on 16 `clippy::manual_is_multiple_of` sites in existing code. The lint is MSRV-gated: `is_multiple_of` is stable since Rust 1.87, so clippy stays silent while `rust-version` says 1.85.0 and fires once it says 1.93. The local clippy run had been taken right after `cargo update` and before the `rust-version` edit, and this body claimed the later manifest edits did not matter because they left the lock unchanged. That was true of the lock and false of the lints.

The sites are converted to `is_multiple_of` rather than holding `rust-version` down, since 1.93 is the floor the graph really needs. Every divisor is non-zero (constants, `progress_step` is at least 5, the delta `grid` is a multiple of `DELTA_PART_SIZE`), so the rewrite cannot change a division-by-zero panic into an answer.

## Verified locally on this head

- `cargo clippy --all-targets -- -D warnings` with `rust-version = "1.93"`: rc 0, re-run from `cargo clean -p aeroftp` on this head (and `peer-l0`: rc 0).
- `tsc --noEmit` (7.0.2): rc 0.
- `vitest run`: 103 files, 910 tests passed.
- `npm ci` with npm 10.8.2: ok, `npm audit`: 0 vulnerabilities. `cargo audit` on the base commit: 0 vulnerabilities.

The full green is CI: this touches the lockfile of every platform build.

## Summary by CodeRabbit

- **Updates**
  - Updated notification support to version 2.4.0.
  - Updated the application’s Rust toolchain requirements to align with current dependencies.
  - Updated the peer component’s minimum Rust version to match its networking dependencies.
- **Maintenance**
  - Improved compatibility with newer package versions.
  - Refined internal progress reporting and validation checks without changing user-visible behavior.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -104,11 +104,6 @@ updates:
         versions: [">=0.8.0"]
       - dependency-name: "chacha20poly1305"
         versions: [">=0.11.0"]
-      # globset 0.4.20 declares rust-version 1.88; we still advertise
-      # 1.85.0. Unblock when the AeroFTP MSRV moves, or globset ships
-      # a 0.4.x that stays on 1.85.
-      - dependency-name: "globset"
-        versions: [">=0.4.20"]
 
   # npm dependencies
   - package-ecosystem: "npm"
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
         "@tauri-apps/plugin-dialog": "^2.7.3",
         "@tauri-apps/plugin-fs": "^2.5.0",
         "@tauri-apps/plugin-log": "^2.9.1",
-        "@tauri-apps/plugin-notification": "^2.3.3",
+        "@tauri-apps/plugin-notification": "^2.4.0",
         "@tauri-apps/plugin-shell": "^2.3.6",
         "@xterm/addon-fit": "^0.11.0",
         "@xterm/addon-web-links": "^0.12.0",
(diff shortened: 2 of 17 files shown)


axpnet:
@coderabbitai full review

PR #780, opened 2026-09-10
chore(deps): take the argon2 0.6, sevenz-rust2 0.22 and zstd 0.14 majors

axpnet:
Takes three of the four Dependabot majors opened today in one pull request, as with #760: [#777](https://github.com/axpdev-lab/aeroftp/pull/777) argon2 0.5 to 0.6, [#778](https://github.com/axpdev-lab/aeroftp/pull/778) sevenz-rust2 0.21 to 0.22, [#779](https://github.com/axpdev-lab/aeroftp/pull/779) zstd 0.13 to 0.14. All three touch only the manifest line and the lockfile, and none needed a source change. The fourth, [#776](https://github.com/axpdev-lab/aeroftp/pull/776) vitest 5, is not in here: see the last section.

## What each bump changes, read from the crates

- **argon2 0.6.0.** The API AeroFTP uses is unchanged: `Params::new`, `Argon2::new(Algorithm::Argon2id, Version::V0x13, params)` and `hash_password_into` keep their signatures in `crypto.rs`, `user_crypto.rs` and `aerocrypt`. The release moves to Rust 2024 and the digest 0.11 / blake2 0.11 cohort, detects allocation failures in `hash_password_into`, and reorders one parameter check. `0.6.0` was already in `Cargo.lock` as a transitive dependency since #773, so the lock gains nothing new; `argon2` 0.5.3 stays for `aerovault` 0.6.4, which still requires `^0.5`.
- **sevenz-rust2 0.22.2** with `lzma-rust2` 0.20.1. Additive: coder properties exposed, a solid pack compressed away from the writer, LZMA2 chunks decoded as they arrive. The one behaviour change, a truncated LZMA2/XZ stream now reporting `UnexpectedEof` instead of `InvalidData`, reaches no decision here: the only `ErrorKind::InvalidData` in the archive paths is the one `copy_entry_bounded` constructs itself for an entry that expands past its declared size.
- **zstd 0.14.0** with `zstd-safe` 8.0.0. `stream::encode_all`, `decode_all`, `copy_encode`, `Decoder`, and the `zstd_safe::{DCtx, InBuffer, OutBuffer}` used by the aerorsync literal decoder all compile unchanged. `zstd` 0.13.3 stays in the graph for `aerovault` and `zip` 8.6, but both versions link the same `zstd-sys` 2.1.0 (libzstd 1.5.7), so there is one C library and no `links` conflict, and frames written by either decode with the other.

## Verified locally on this head

- `cargo check --all-targets`: rc 0.
- `cargo test --lib -- aerocrypt user_crypto crypto keystore_export compress_overlay archive 7z sevenz zstd real_wire`: 408 passed, 1 ignored. This includes `audited_profile_is_128mib_t4_p4_and_derives_the_known_answer` and `argon2_and_hkdf_are_deterministic`, which would fail if the argon2 bump changed a single derived byte, plus the zstd round-trip and decompression-bomb tests and the aerorsync zstd literal stream tests.
- `cargo clippy --all-targets -- -D warnings`: rc 0.

## Why vitest 5 is not here

`vitest` 5.0.0 declares `engines.node` as `^22.12.0 || ^24.0.0 || >=26.0.0`, and every workflow that runs it installs Node 20 (`checks.yml`, `build.yml`). Taking it means moving CI off Node 20 first, which is past end of life and worth doing on its own, but it changes the npm version CI installs with and therefore the lockfile convention (npm 10.8.2), so it belongs in a change of its own rather than inside a dependency batch. #776 stays open until that is decided.

## Summary by CodeRabbit

* **Chores**
  * Updated internal components supporting password protection, archive handling, and compression.
  * No user-facing features or behavior changes are included.
  * Existing functionality and configuration remain unchanged.


axpnet:
@coderabbitai full review

PR #781, opened 2026-09-10
ci: move from Node 20 to Node 24 LTS and take vitest 5

axpnet:
Moves every CI workflow that installs Node from Node 20, which reached end of life in April 2026, to Node 24, the current Active LTS ("Krypton"), and takes vitest 5, which Dependabot proposed in #776 and which requires Node 22.12 or later.

## Node 24

| Where | Change |
|---|---|
| `build.yml` (2 jobs), `checks.yml`, `portal-chooser.yml`, `snap-refresh.yml` (2 jobs) | `node-version: '20'` to `'24'`, every occurrence in the tree |
| `CONTRIBUTING.md` | Node.js 24+, CI on Node 24 LTS |

**Deliberately not changed: the Flatpak manifest.** `com.aeroftp.AeroFTP.yml` stays on `org.freedesktop.Sdk.Extension.node20`. The Flatpak packaging is not finished and will be revisited together with the app id rename, so it does not change here in a pull request that cannot build it. When that work resumes, `node24` exists for the freedesktop 25.08 base of its GNOME 49 runtime (24.20.0 on Flathub at the time of writing).

Node 24 over Node 22 because 22 is already in maintenance, while 24 is the Active LTS with support into 2028. Node 26 is not an LTS yet.

**Lockfile convention changes with it.** CI installs with the npm that ships with its Node, and on Node 24 that is npm 11 (11.19.0 with 24.21.0) instead of npm 10.8.2. `package-lock.json` in this PR is generated with `npx npm@11.19.0`, so `npm ci` in CI reads a lock written by the same npm. The lock diff is confined to vitest's own tree: nine packages vitest 5 now inlines (`@vitest/runner`, `@vitest/utils`, `@vitest/expect` and others) go, eleven of its dependencies change version, and no package outside that tree moves.

## vitest 5

The breaking changes in the 5.0.0 release notes that could touch this repository do not: Node 22+ and Vite 6.4+ (we are on Vite 8.2.2), `expect.poll` failing when the function never resolves (not used here), and config lookup no longer walking up ancestor directories (`vitest.config.ts` sits at the repository root). The config is a node environment over `src/**/*.test.ts`, with nothing else configured.

## Verified locally on this head (Node 24.12.0)

- `npm ci` with npm 11.19.0: ok, 0 vulnerabilities.
- `vitest run` (5.0.0): 103 files, 910 tests passed, the same counts as on vitest 4.1.11.
- `tsc --noEmit` (7.0.2): rc 0.
- `npm run build` (`tsc` and `vite build`): ok.

Supersedes #776.

## Summary by CodeRabbit

- **Chores**
  - Updated project build, validation, release, and refresh processes to use Node.js 24.
  - Updated automated checks to run on Node.js 24.
  - Updated testing tools to the latest major version for improved compatibility.

- **Documentation**
  - Updated contributor setup requirements to specify Node.js 24 or later.

- **User Impact**
  - No direct changes to application features or user-facing behavior.


axpnet:
@coderabbitai full review

PR #798, opened 2026-09-11
fix: close the plausible findings from the v4.1.9 audits

axpnet:
## What this is

A sweep of the PLAUSIBLE-grade findings from the v4.1.9 pre-release audits, one commit per finding. Verified against the code of today first; the table at the bottom lists what was confirmed, what was fixed, and what was dismissed.

## Findings

| Finding | Status | Commit |
|---|---|---|
| DCO check matched the sign-off as a substring anywhere in the commit message, so quoting it in prose passed without a trailer | Confirmed; the match now anchors on lines starting with `Signed-off-by:` first | `ci(dco): match the sign-off on trailer lines, not quoted prose` |
| `assert-linux-acl-packaging.sh` set a `RETURN` trap at script top level, which never fires in an executed (not sourced) script | Confirmed; dead trap removed, explicit cleanup kept, matching the .deb and snap branches | `fix(scripts): drop the dead RETURN trap from the AppImage check` |
| The yanked-crate check passed green on a registry timeout: `cargo audit` reports a yanked crate, and an unrun yank check, as a warning with exit 0 | Confirmed; see the policy change below | `ci(security): deny cargo audit warnings so a skipped yank check fails` |
| `Cargo.toml` comment cited "aerovault 0.6.3 depends on 0.10" under the 0.6.4 declaration | Confirmed; one word corrected, the pin itself still holds | `chore(deps): name aerovault 0.6.4 in the chacha20poly1305 pin comment` |
| The CR/LF command-injection refusal existed only as a manual reproduction; the #672 fixture counters were never consumed | Confirmed; new `ftp_crlf_refusal.rs` drives upload and delete with a CR/LF name against a recording loopback server and asserts the refusal is typed `InvalidPath` and the injected verb never crosses the wire | `test(ftp): pin the CR/LF command-injection refusal on the wire` |
| Em-dash in `rsync_over_ssh.rs` and eight in `jottacloud.rs` | Confirmed; replaced with colon or parentheses | `style(aerorsync): ...`, `style(jottacloud): ...` |
| IntroHub reorder sentinels captured external drags: the shared dragenter/dragover/drop handlers called `preventDefault()` and set `dropEffect` unconditionally | Confirmed; the handlers now claim the drag only while the panel's own reorder is in progress (synchronous ref, not `dataTransfer.types`, because the internal drag publishes plain text). What changes for the user: the reorder drop targets no longer accept external drops; an external file or text dragged over the server list no longer lights the targets up. Declared limit: no live GUI test exists for this; the project has no React component test harness, so the change is covered by typecheck and the unit suite only | `fix(introhub): claim only internal reorder drags on drop targets` |

One fix emerged while writing the CR/LF test: `delete` and `rmdir` mapped the library's unwritable-command-line refusal to `ServerError`, so a crafted name on those paths surfaced as "Connection error" and read as retryable by sync's substring classifier, the exact mistyping the transfer paths had already fixed. Both now apply the same narrow io-kind rule via `classify_command_refusal`. The delete test was red before that commit and green after.

## CI policy change: `cargo audit --deny warnings`

`checks.yml` and the local pre-push gate now run `cargo audit --deny warnings`. This is a policy change: from now on a new warning (for example a crate newly declared unmaintained, or a yanked release) and a registry timeout that leaves the yank check unrun both fail red on the authoritative lanes (main pushes, weekly schedule, local gate). The PR lane keeps `continue-on-error`, so a new upstream advisory does not red-wall unrelated pull requests. The dependency tree is warning-clean today (measured: 1243 advisories loaded, 1206 crate dependencies scanned, no warnings).

Remedies when it fires: an accepted warning is justified in the commit and added to the ignore list in `src-tauri/.cargo/audit.toml`; a registry timeout is simply re-run.

## Dismissed or deferred

- The two other em-dashes named in the audit (`rclone_crypt.rs`, `agent_session.rs`) are already gone on current main; nothing to do.
- The dead pCloud trash path noted in the same audit (unreachable `PCloudTrashManager` still wired to live commands) was reported in triage and is deliberately left out of this sweep; it is its own removal.

## Gates

`cargo fmt --all` clean; `cargo test --test ftp_crlf_refusal` 2/2; `cargo test --lib ftp` 125/125 and `--lib jottacloud` 33/33; `cargo clippy --all-targets -- -D warnings` clean; typecheck, i18n validation, provider inventory and the unit suite (916 tests) all green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved FTP error handling: paths containing invalid command characters are now rejected clearly as invalid paths.
  * Prevented accidental handling of external text or file drags in the server reordering area.
  * Strengthened validation for potentially unsafe FTP path inputs.

* **Quality Improvements**
  * Enhanced automated checks to catch security and validation issues earlier.

## Update (2026-09-12)

The IntroHub change is out of this pull request. `80e8a2ddb` guarded `dragenter`, `dragover` and `drop` on the panel's own reorder, and in doing so it removed the unconditional `preventDefault()`. The app disables the native Tauri drag handler, so on Windows WebView2 navigates away from the app when a file is dropped on a card or on one of the sentinels. `eeada1b85` undoes it, and `MyServersPanel.tsx` is now identical to the version on `main`. Claiming only internal reorders is still wanted: it returns as its own change, with a capture-phase guard that keeps external files cancelled, and a check on Windows.

### DIFF
--- a/.github/scripts/pre-push-gate.sh
+++ b/.github/scripts/pre-push-gate.sh
@@ -47,5 +47,7 @@ cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets -- -D warnings
 cargo test --manifest-path src-tauri/Cargo.toml
 (
     cd src-tauri
-    cargo audit
+    # Same strictness as checks.yml: a yanked crate or an unrun yank check is
+    # a warning with exit 0 by default, which reads as a pass. Deny them.
+    cargo audit --deny warnings
 )
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -150,5 +150,11 @@ jobs:
         # Non-blocking on PRs: a newly published advisory on a transitive
         # dependency must not red-wall an unrelated PR. On push to main and on
         # the weekly schedule the step is authoritative and fails the job.
+        # `--deny warnings`: without it a yanked crate, and a registry timeout
+        # that leaves the yank check unrun, both end as a printed warning with
+        # exit 0, indistinguishable from a pass (#661 declared that gap; the
+        # v4.1.9 audit measured it still open). The dependency tree is
+        # warning-clean today, so an unavailable check now fails red here
+        # instead of being read as clean.
         continue-on-error: ${{ github.event_name == 'pull_request' }}
-        run: cargo audit
+        run: cargo audit --deny warnings
--- a/.github/workflows/dco.yml
+++ b/.github/workflows/dco.yml
@@ -53,7 +53,9 @@ jobs:
                 shown="$expected"
                 ;;
             esac
-            if git show -s --format='%B' "$sha" | grep -qiF "$expected"; then
+            # Anchor on real sign-off LINES before the substring match: a commit
+            # that quotes "Signed-off-by: ..." inside prose must not pass.
+            if git show -s --format='%B' "$sha" | grep -i '^Signed-off-by:' | grep -qiF "$expected"; then
               continue
             fi
             echo "::error::commit $sha is missing '$shown'"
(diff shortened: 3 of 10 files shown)


axpnet:
@coderabbitai full review

PR #810, opened 2026-09-12
fix(mcp): say how much of each tree check_tree actually compared

axpnet:
## The defect

`aeroftp_check_tree` compared two trees and never said how much of them it had actually read.

It called `scan_local_tree` and `scan_remote_tree`, the wrappers whose whole body is `scan_*_checked(...).0`: they drop the `ScanCompleteness` of the walk and the `ScanBoundaries` it produced. So a directory that failed to list, and a directory behind a symbolic link that no walk follows, both came back as nothing at all. In the result they are indistinguishable from a directory that is genuinely absent on the other side.

That distinction is the entire question when the agent's next step is a delete: a file "missing" from an incomplete scan may simply be unread.

This was named as a sibling and left out of scope by the change that fixed the same class in the CLI `check` command, with a note that it would be reported separately. This is that report.

## Fix

The tool now uses the checked variants and reports what they return, in the shape `aeroftp_sync_tree` already established rather than a new one:

- `{side}_skipped_links` and `{side}_unseen_paths`, emitted through the same `insert_capped_list` helper, capped at 1000 entries with `_total` and `_truncated` beside them, and absent entirely when empty.
- `{side}_scan_incomplete`, `{side}_scan_errors`, `{side}_scan_truncated` for each of the two sides.
- Under `summary_only=true` only the counters go out, not the path lists. That mode exists to keep a response inside the MCP size limit, and two lists of a thousand paths would defeat it.

### Two decisions worth stating

**`{side}_scan_incomplete` is not `ScanCompleteness::is_complete()`.** A link is a gap without being a failure: the subtree behind it was never compared, while `list_errors` stays 0 and `truncated` stays false. A field that asked only the counters would report a complete scan in the same object that lists the path it did not cover. That exact contradiction was found and fixed in the CLI `check` report earlier today, and it would have shipped here unchanged.

**The six counter keys are always present, zero included**, while only the path lists come and go. A caller that has to check whether a key exists before reading it will not check.

## Documentation

The tool description is the contract an agent reads before calling, and in this codebase descriptions are pinned by tests, so it is part of the change rather than a comment on it. It now names the keys, says which are always present, says what `summary_only` drops, and ends with the sentence that matters: a file reported as missing by an incomplete scan may be unread rather than absent, so it must not be turned into a delete.

## Tests

Three, and all three are guards rather than red-then-green: these keys did not exist, so there was nothing to be red. That is stated here rather than left for the reader to work out.

- `check_tree_result_counts_a_link_as_an_incomplete_scan` pins the term that would be dropped first by anyone simplifying the helper back into `is_complete()`, and also pins that the link is not reported as a truncation and that the clean side is not tarred with the other side's gap.
- `check_tree_result_always_carries_the_counters_and_drops_the_lists` pins the presence rule in both directions, including under `summary_only`.
- `check_tree_description_documents_the_scan_coverage_keys` pins the contract in the description.

```
mcp::tools::tests::check_tree     6 passed; 0 failed
mcp::                            57 passed; 0 failed
```

## Gates

- `cargo fmt --all -- --check`: rc 0
- `cargo test --lib mcp::`: 57 passed, 0 failed
- `cargo clippy --lib --tests -- -D warnings`: rc 0, run after `cargo clean -p aeroftp` so the lint pass recompiled the crate instead of answering from cache
- Frontend gates were not re-run and are not required by this head: one Rust module, with no TypeScript, locale or provider catalog touched.

No CHANGELOG entry: the release notes are written from the tracker at release time.

### DIFF
--- a/src-tauri/src/mcp/tools.rs
+++ b/src-tauri/src/mcp/tools.rs
@@ -93,7 +93,7 @@ pub fn tool_definitions() -> Vec<McpToolDef> {
         },
         McpToolDef {
             name: "aeroftp_check_tree",
-            description: "Compare a local directory against a remote directory and report differences...",
+            description: "Compare a local directory against a remote directory and report differences. The result also says how much of each tree the comparison actually covers: `local_scan_incomplete` and `remote_scan_incomplete`, with `{side}_scan_errors` (directory listings that failed) and `{side}_scan_truncated` (the walk stopped at its depth or entry cap). A side is reported incomplete when its counters are non-zero OR when the walk left out a path it can name, because a symbolic link is a gap without being a failure: no walk follows one, so the subtree behind it was never compared. Those paths are listed in `{side}_skipped_links` and `{side}_unseen_paths`, each capped at 1000 entries with `_total` and `_truncated` alongside, and OMITTED entirely when empty. The six counter keys are always present, zero included, so an agent can branch on them without first checking whether they exist. Under summary_only=true only the counters and flags are returned, not the path lists. A file reported as missing on one side by an incomplete scan may simply be unread rather than absent: do not turn it into a delete.",
             input_schema: json!({ "type": "object", "properties": {
                 "server": { "type": "string", "description": "Server name or ID" },
                 "local_dir": { "type": "string", "description": "Local directory to compare" },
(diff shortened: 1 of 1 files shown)


axpnet:
@coderabbitai full review

PR #811, opened 2026-09-12
fix(sync): take a skipped link type from its target

axpnet:
## The defect

`SkippedLink.is_dir` was a `bool`, and two of the places that filled it asserted a value they had never observed.

**The local prefix in `ScanBound::for_sync`.** A comment there explained that a local link pointing at a FILE could not reach that loop, because such a link would be listed as a file and therefore sit in `local_set`. That is false. `scan_local_tree_checked` skips EVERY symlink, links to files included (the `Ok(_) => {}` arm), so the path never enters `local_set`, the loop offers the whole path, and the link was recorded as standing where a directory does. An independent review reproduced it on a real filesystem: a remote entry `leaf` plus a local `leaf -> target.txt`.

The comment is the part worth naming. It did not merely fail to prevent the defect: it argued that the defect was impossible, which is the kind of sentence that stops the next reader from checking.

**The walk-error arm.** It recorded a directory whenever the error carried a path. That arm is reached for a listing that failed, where the path is indeed a directory, and also for a directory entry that could not be built: on a filesystem that needs a stat to type a child, that child can be a file. One arm, two shapes.

## Why it mattered, and why it did not yet hurt

The type is read by the crypt boundary decoder, which spells a directory differently from a file: with `Off` mode and a suffix, a directory carries none, and with `directory_name_encryption` off a directory name stays in the clear. A wrong type there anchors a bound to a path that does not exist, silently.

Today that decoder only sees REMOTE boundaries, and both defects are on the local side, so the harm was contained. That containment is a property of the current call graph, not of the code being right: the day a local boundary reaches a decoder, these become live. Fixing them now costs a type change; fixing them then would cost an investigation.

## Fix

- `is_dir` becomes `Option<bool>`, the shape `UnseenPath` already used, so "unknown" is expressible rather than rounded to a guess.
- The `for_sync` prefix reads the type from `metadata`, which follows the link, and records `None` when the target does not resolve. The false comment is gone, replaced by what was actually measured.
- The walk-error arm records an unknown type.
- The three sites where the type IS observed keep their value, each with one line saying what observed it: the flat listing that typed the entry, the arm that matched on the target being a directory, and the compare that matched on `is_dir && is_symlink`.
- The crypt boundary decoder stops wrapping a link's value in `Some`: it takes the option directly and already refuses a path whose type is unknown instead of guessing.

The compiler enumerated the sixteen test literals the type change reached, across four files. Each was rewritten from the value it already carried rather than from a guess, which matters because two of them were `false` and had to stay false.

## Tests

Three, all on the behaviour that changed.

### Seen red, and how

The fix and the type change cannot be split into a commit that still compiles, so the red was obtained by temporarily restoring the old assertion at the two sites, running the three tests, and restoring the file from a copy taken beforehand. The method is stated here rather than implied by a red-then-green commit pair that did not happen:

```
link_bound_types_a_local_link_to_a_file_from_its_target
  left: [SkippedLink { rel_path: "leaf", link_target: Some("target.txt"), is_dir: Some(true) }]
 right: [SkippedLink { rel_path: "leaf", link_target: Some("target.txt"), is_dir: Some(false) }]

link_bound_leaves_the_type_unknown_when_the_target_does_not_resolve
  left: [SkippedLink { rel_path: "dangling", link_target: Some("nowhere"), is_dir: Some(true) }]
 right: [SkippedLink { rel_path: "dangling", link_target: Some("nowhere"), is_dir: None }]

scan_local_tree_leaves_the_type_unknown_for_a_path_it_could_not_open
  left: [("locked", Some(true))]
 right: [("locked", None)]
```

The restoration was verified before the green run, by asserting that the `metadata` line was back and that no stub remained.

### Green after the fix

```
sync_core::scan::tests      45 passed; 0 failed
provider_commands::tests    66 passed; 0 failed
mcp::tools::tests           38 passed; 0 failed
transfer_dag_sync::tests    35 passed; 0 failed
```

### Declared coverage limit

The test for the walk-error arm needs a directory the user cannot open, so it is `#[cfg(unix)]` and skips itself under root, saying why: the existing helper in that module asserts that the mode really blocked, so the test cannot pass without observing its case.

## Gates

- `cargo fmt --all -- --check`: rc 0
- `cargo test --lib` on the four modules above: 184 passed, 0 failed
- `cargo clippy --lib --tests -- -D warnings`: rc 0, run after `cargo clean -p aeroftp` so the lint pass recompiled the crate instead of answering from cache
- Frontend gates were not re-run and are not required by this head: Rust only, with no TypeScript, locale or provider catalog touched.

No CHANGELOG entry: the release notes are written from the tracker at release time.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved synchronization handling for symbolic links whose targets are files or directories.
  - Unresolved link targets are now handled safely instead of being treated as directories by default.
  - Remote comparisons no longer proceed when a link’s type cannot be determined, helping prevent incorrect synchronization results.


axpnet:
@coderabbitai full review

PR #848, opened 2026-09-17
fix(cli,github): delta download in sync, size exit code, error labels

axpnet:
Three fixes the owner decided together on 2026-09-17, kept in one pull request because the repository squashes and three one-file branches would cost three CI cycles for two hours of real work.

## 1. `sync --delta --direction download` moved whole files (commit `3a3e7452`)

The delta engine has always routed both directions: `try_delta_transfer_with_batch` dispatches `SyncDirection::Download` to `DeltaBatch::download`, `AerorsyncBatch::download_file` implements it with the same retry budget as its upload twin, and `cmd_get` has used it in production for single files. The CLI sync loop was the one caller that never asked for it, so the same 8 MiB file with 4096 changed bytes cost 8754570 bytes through `sync --delta --direction download` and 27036 bytes through `get --delta`, with exit 0 and no warning. The message printed for `get --recursive --delta` points the user at `sync --delta`, which is what made the gap easy to miss: the sentence is true for uploads and was misleading for downloads.

The other recursive callers already ask for it, though it is worth naming them precisely because "AeroSync" is ambiguous here. `sync_tree_core` passes the batch to `perform_download`, which uses it with `SyncDirection::Download`, and its only production caller is the MCP surface; in the DAG the batch is opened with no direction gate while `perform_sync_transfer` routes by `transfer.op`. The desktop sync runner does not go through either: it calls `provider_download_file` with `useDelta`, which is the single-shot delta path without a batch, symmetric between upload and download, and the mirror flows turn delta off in both directions. The background cloud sync uses no delta at all, in either direction. So there was one asymmetric gap and it was in the CLI, which is also where it was measured.

Eligibility is deliberately narrower than on the upload side. A delta download reconstructs from a local base, so only jobs whose relative path already exists locally go through the batch; a file absent locally would travel in full anyway plus the cost of a signature exchange, and it goes straight to the classic path with the log saying so rather than claiming a delta that did not happen.

Measured on a live SFTP server (rsync 3.2.7 over sshd), 8 MiB with 4096 changed bytes, local copy older than the remote one:

| | `bytes_on_wire` reported | bytes the server actually sent | resulting file |
|---|---|---|---|
| this branch | 23303 | 18829 | byte-identical |
| released 4.1.9 | no delta path taken | 8428937 | byte-identical |

The second column is read from the container's own interface counters, which know nothing about AeroFTP, so the claim rests on two independent probes rather than on the application's own log. Both runs end with the same SHA-256: the result was always correct, the cost was not.

Eligibility cases: with one file that has a local base next to one that does not, the plan is 2 downloads and the log reports 1 with a local base, both arriving; with an empty local directory, no batch is opened and the fallback is printed.

## 2. `size` exited 0 on a cancelled scan (commit `c93e5f29`)

A cancelled scan prints a figure that is a lower bound, and it exited 0, so a script could not tell an interrupted measurement from a complete one: the number was believable and wrong. It now exits 4, which this CLI already publishes as "transfer failed / partial" and which `reconcile` uses for exactly this meaning, so nothing new enters the public contract of exit codes. The other lower-bound cases (`hit_cap`, `truncated`, `unreadable_dirs`) keep exiting 0: they are declared in the text and in the JSON, and widening an exit code as a side effect of this change would be a second decision nobody took.

## 3. GitHub errors wrote their label twice (commit `059aaa2d`)

`GitHubError` has a hand-written `Display` whose arms add a prefix of their own, and `ProviderError` derives its text with `thiserror`, which adds one more. Every GitHub not-found therefore read `Path not found: Not found: ...` and every parse failure `Parse error: Parse error: ...`. The duplication comes from the pair of types and not from the payload, which is why fixing call sites never removed it.

Six arms collide and only those are stripped, in the conversion alone: `PermissionDenied`, `PathNotFound`, `NotFound`, `NetworkError`, `ServerError`, `ParseError`. The arms that carry a sentence of their own (`Branch 'x' does not exist. Check the branch name or create it first.`, `Release 'x' not found.`, `Release asset 'x' already exists.`) read correctly under their label and are left untouched. The `Display` of `GitHubError` is not changed either, because it is what the logs print and it is read without a `ProviderError` label in front of it.

The guard that had to be protected: `message_names_a_missing_path` is the one door that turns an error into an absence, and it looks for the not-found vocabulary. The label the user reads still comes from `ProviderError`, so `Path not found: <payload>` keeps those words, and a test asserts that an absence is still recognised after the strip.

## Tests

Both new behaviours were seen red before they were seen green. `provider_error_does_not_repeat_the_label_of_the_variant` fails on the unfixed conversion (`Path not found: Not found: gone.txt` counts the label twice), and `size_of_a_cancelled_scan_exits_4_like_reconcile` fails against the previous return value. The three other GitHub tests are guards rather than falsifiers, and their comments say so.

`cargo fmt --check` is clean. `cargo check --bin aeroftp-cli --features aerorsync` is green, verified with a positive control: a deliberate type error inserted in the new function made the check fail on that line, so the green describes this tree and not a cached one.

Not covered here: Clippy and the full suite are left to CI, `--direction both` opens one batch per phase (declared in the log, not exercised live), and the Windows live run is being done separately, where `aerorsync` is first-class rather than the preferred path.

## Follow-ups from review, in this same branch

A local reviewer went through the branch before CI finished and found four things worth fixing here rather than later.

**`--max-transfer` was counted but not enforced on the delta leg** (`5412e4da`, corrected by `f2fdc534`). The new branch called the session counter but never checked the budget, so a run with a small budget let every eligible file through the delta leg, spent the budget there, and then truncated the classic queue the flag was meant to protect. The first attempt stopped the leg by pushing an error, and that was the wrong verb: `cmd_sync` reads a non-empty error list before it reads the cap, so a deliberate ceiling exited 4 instead of 8, and 4 is what the agent guide defines as a retryable failure. A tool would have re-run a sync that did exactly what it was told, and files the classic path reports as skipped with an honest note would have been listed as failures.

The remainder is now handed back to the classic queue, whose pre-flight refuses it with that note and moves no bytes, because `cap_files_to_max_transfer` starts over the limit and breaks on the first file. One rule, applied in one place. The bytes counted are those of the reconstructed file read from local metadata, which is what `cmd_get --delta` and the classic task both do: the remote listing figure is the one number that can be stale.

Measured live, red before green: two 4 MiB files with a local base and `--max-transfer 1M` exited 4 with `Error: download a.bin: max-transfer limit reached`; after the correction they exit 8 with an empty error list and `Note: --max-transfer reached, skipped 1 file(s)`.

**The `size` JSON contradicted its own exit code** (`4af01a97`). It exits 4 on a cancelled scan and still said `"status": "ok"`, so a reader who trusts the JSON learned the opposite of a reader who trusts the exit code. It now says `"partial"`, the word `reconcile` uses for the same outcome. The public CLI guide still documented the old behaviour and is corrected in the same commit, including the note that `df --scan` shares this engine and this signal but still exits 0 when cancelled.

**The doubled-label fix covered one door of two** (`c5538045`). Eleven sites built a `ProviderError` from a `GitHubError` by hand rather than through the conversion, so a 502 on the release list still read `Server error: GitHub server error: ...`. They now go through `From`, which also classifies them properly: a 404 on a release is a `NotFound` and no longer a `ServerError`. The sites wrapping `reqwest` or `serde_json` are untouched, because there is no doubled label to remove there.

**The strip match no longer has a wildcard arm.** It lists every variant, so a variant added later that collides with a `ProviderError` label breaks the build instead of being absorbed in silence, which is how the doubling survived unnoticed the first time. The absence guard now asserts the variant as well as the words: the words come from the `thiserror` label and would survive almost any edit, while routing those arms to `Other` or `ServerError` is the change that would really lose an absence.

## Declared limits

These are known and not addressed here, listed so they are not discovered as surprises.

The delta leg ignores `--bwlimit` and `--limit-rate`, `--parallel`, and `--inplace`/`--partial`; it does not advance the overall progress bar and does not appear in `engine_stats` in the JSON summary. Ctrl-C does not call `batch.cancel()`, the same as its upload twin. The delta writer sets remote mtime and mode while the classic path does not, so one run can leave two files with different metadata. `--direction both` opens one batch per phase, which is two SSH sessions, declared in the log and not exercised live. The eligibility filter is proved by live runs (with a base, without a base, mixed) rather than by an automated test, and the `size` test fixes the mapping rather than the wiring from `s.cancelled`, which needs a live connection to exercise.

## Two things looked at and deliberately left alone

Routing the eleven hand-built errors through `From` also changes how they are classified, and that is not cosmetic: `get_raw` now yields `NetworkError` where it used to yield `TransferFailed`, and a 404 on the release list becomes `NotFound`. GitHub answers 404 for a private repository without access as well, so the question is whether a generic consumer could read a `NotFound` on a release listing as "the remote is empty". Repo mode already went through `From` for the same 404, so this makes the two modes agree rather than introducing a new shape, and the absence door reads the final text of a `ProviderError` rather than the listing itself. Worth knowing when reading a bug report about an empty release list.

Measuring the cap surfaced a defect that belongs to neither of these three voices and is left for its own fix: when `--max-transfer` is reached, the sync JSON still reports `"status": "ok"` and `"skipped": 0` while the process exits 8 and a file really was skipped. The released 4.1.9 binary, with no `--delta` at all, produces the same three values on the same bench, so this predates the branch. It is the same class as the `size` mismatch fixed here, on a different command.

## Summary by CodeRabbit

- **New Features**
  - `size` now reports partial results with status `partial` and exits with code 4 when a recursive scan is cancelled.
  - `sync --delta` enforces the configured maximum transfer limit and accurately accounts for reconstructed-file sizes.

- **Bug Fixes**
  - Improved GitHub error reporting by removing duplicated labels while preserving meaningful details and missing-path detection.
  - GitHub operations now retain more specific error classifications.

- **Documentation**
  - Clarified cancellation behavior for `size` and distinguished it from `df --scan`.
  - Clarified that delta transfers work with password-only SFTP servers.


axpnet:
@coderabbitai full review

PR #867, opened 2026-09-19
test(dispatcher): a unique directory per test, not a unique nanosecond

axpnet:
Closes the CLI Smoke `coverage` red on `main` (run 35381974740 on `ed07b6347`): `dispatch_execs_gui_stub_with_linux_webkit_env` panicking with `dispatcher exec failed after 2 attempt(s): No such file or directory (os error 2)`.

There turned out to be **two** defects behind that job, with different signatures. Both predate this branch.

## Reading the message before explaining it

The retry loop in `run_dispatcher` continued on ETXTBSY and on nothing else, so "after 2 attempt(s)" with an ENOENT meant: attempt 1 hit ETXTBSY, slept 20 ms, attempt 2 found the file gone. Not an isolated ENOENT, and that shape is what pointed at something deleting the file mid-run rather than at a missing interpreter.

## Defect 1: two tests sharing one directory

`TestDir::new` built its name from the pid and the nanosecond. All three tests here used that same template, so two of them reading the same nanosecond got the **same directory**, copied the dispatcher to the **same path**, and the rest followed: one test exec'd a file the sibling still held open for writing (ETXTBSY), the sibling finished, its `Drop` ran `remove_dir_all`, and the retry found nothing (ENOENT).

Fixed with a process-wide `AtomicU64` in the name, so the directory is unique **by construction** rather than by clock resolution. Not another retry: a retry would have waited out the overlap instead of preventing it, and the sibling would still have deleted the other's directory.

## Defect 2: the dispatcher's own exec of the stub

With defect 1 closed, a second red remained at a much lower rate and with a different signature: exit 127 where the stub's own code was expected. 127 is the dispatcher's error branch, so the failure was not on its exec (the retry never fired) but on its single `execve` of the stub, inside another binary where no retry of ours reaches. Captured stderr: `aeroftp-dispatch: exec failed: Text file busy (os error 26)`.

The cause is ordinary and unrelated to the directories: the kernel refuses to exec a file open for writing anywhere, and `O_CLOEXEC` closes a descriptor at the child's exec, not at the fork. Between a sibling's fork and its exec, the child holds a copy of every writable descriptor this process had open, so a stub being written right then is busy for as long as that lasts.

Fixed with `FORK_GUARD`, an `RwLock` that serialises "an executable is open for writing here" against "this process forks": writers hold it exclusively across open, write, chmod and close, anything that spawns holds it shared. Execs are never serialised against each other, which is what the tests measure, and descriptors are per process, so this process guarding itself is the whole fix. **The ETXTBSY retry is removed with it**: it could only ever have covered the exec of the dispatcher, not the dispatcher's exec of the stub where the failure actually landed, and keeping it would leave two explanations in one file for a window that no longer exists.

## No product change, and why

`ETXTBSY` at launch would need an install that rewrites `aeroftp.bin` in place. It was checked rather than assumed: `dpkg` and `rpm` unpack to a temporary name and rename, so the running inode is never opened for writing, Snap serves read-only revisions, and the portable ZIP with its in-place updater is **Windows only**, where the dispatcher goes through `CreateProcess` and `ETXTBSY` does not exist. So the dispatcher keeps its single `execve`.

## Measurements

Every number below was measured on a Linux workstation, the reds under concurrent load from another branch's full compile.

| # | Scenario | Result |
|---|---|---|
| 1 | Can two `TestDir` names collide: three threads sampling the clock together, as the three tests do | **21 collisions in 200000 rounds**, smallest observed gap **0 ns** |
| 2 | Is there any other code that could delete the dispatcher mid-run | One line: `TestDir::drop`. Nothing else in the file removes anything |
| 3 | The guard test's **first** form, asserting only that 64 paths were distinct | Would have caught a return to a clock-only name in **0.3% of batches** (7 of 2000): decoration, so it was rewritten |
| 4 | The guard test's **second** form, asserting the draws were consecutive | **8 red runs in 200** at `--test-threads=8`: flaky for the very reason this branch exists, so it was rewritten again |
| 5 | The guard test as it stands, asserting the draws are distinct, against the OLD name | **Red on every run**: 1 distinct value against 64 expected |
| 6 | Defect 2's rate before `FORK_GUARD` | **3 reds in 6200 runs** under load. Two were captured and both read `exec failed: Text file busy`; the third is counted although its output was not saved |
| 7 | After `FORK_GUARD` | **0 reds in 3000 runs** under load |
| 8 | Does poisoning hide the first failure: a writer panicking while holding the guard | `read().unwrap()` panics with `PoisonError`; `read().unwrap_or_else(\|e\| e.into_inner())` proceeds. Both sides now use the latter, so the panic that poisoned the lock stays the only red |

The guard test took three attempts and the first two are recorded in its comment rather than tidied away, because a guard that fires three times in a thousand while its comment claims otherwise is worse than no test at all.


axpnet:
@coderabbitai full review

PR #868, opened 2026-09-19
build(flatpak): move the SDK extension from Node 20 to Node 24

axpnet:
Two lines in `app.aeroftp.AeroFTP.yml`: `org.freedesktop.Sdk.Extension.node20` becomes `node24`, and `/usr/lib/sdk/node20/bin` in `append-path` becomes `/usr/lib/sdk/node24/bin`, matching the Node 24 every CI workflow uses since #781.

It was held back only to avoid a rename/modify conflict with the identifier rename, which landed in #814, so it now applies to the renamed file directly.

Verification: checked against Flathub, not built. `org.freedesktop.Sdk.Extension.node24` is published for branch 25.08 (24.21.0), the freedesktop base of the GNOME 49 runtime this manifest declares. The Flatpak is not a release channel today and no CI job builds it.

## Summary by CodeRabbit

* **Chores**
  * Updated the bundled Node.js runtime environment to version 24.


axpnet:
@coderabbitai full review

PR #869, opened 2026-09-19
fix(cli): correct three help and banner texts, skip the S3 delta on Filebase

axpnet:
Four small corrections found by the v4.2.0 pre-release documentation audit, all in text or behaviour of the CLI itself rather than in the docs.

| What | Before | After |
|---|---|---|
| `--sftp-download-preset` help | "`efficient` matches the GUI default" | "`fast` (eight connections) matches the GUI default" (true since #768) |
| `--help` exit-code table | the `99 / 130` row printed after the exit 8 paragraph | back inside the table |
| Banner | "23 providers" written by hand | protocols and native integrations read from `docs/PROVIDER-INVENTORY.json` (generated, kept in step with the registry by CI) |
| `put --delta` to Filebase | opened a multipart, took 501 on `UploadPartCopy`, aborted, fell back, on every invocation | refused up front with `range_copy_unsupported`, next to the Filen gate |

## Tests

- `banner_counts_come_from_the_generated_inventory`: the banner falls back to zero instead of failing, so this test is what makes a moved file or a renamed key visible.
- `filebase_refuses_a_delta_before_any_request`: the provider points at `s3.filebase.io` with no server behind it, so an `Ok(Refused("range_copy_unsupported"))` proves the gate answered before any request.
- Local: `cargo fmt --check` rc 0; `cargo test --lib filebase_refuses_a_delta` 1 of 1; `cargo test --bin aeroftp-cli inventory_tests` 9 of 9.

## Summary by CodeRabbit

- **New Features**
  - Startup banners now display up-to-date provider counts, including transport protocols and native integrations.
  - CLI help now presents exit codes 99 and 130 within the exit-code table.
  - Updated SFTP preset guidance clarifies that “fast” uses eight connections, matching the GUI default.

- **Bug Fixes**
  - Delta uploads to Filebase S3 endpoints are now refused cleanly before any unsupported request is attempted.


axpnet:
@coderabbitai full review

axpnet:
@coderabbitai full review

Issue #322, opened 2026-06-13
[BUG] AeroFile doesn't decrypt `.cryptomator` files


EhudKirsh:
**Describe the bug**
As the title says. It returns this error at the top of the Cryptomator Vault window in red font:
"Failed to read masterkey.cryptomator: The system cannot find the file specified. (os error 2)".

**Steps To Reproduce**
1. Go to AeroFile 📁
2. Right-click on a `masterkey.cryptomator` file.
3. Click on 'Open as Cryptomator Vault'.
4. Select any folder.
5. Enter the password in the Vault Password field.
6. Click on '🔓 Unlock Vault'.
7. See the error at the top of the window.

**Expected behavior**
To decrypt `.cryptomator` files.

**Environment:**
 - OS: Windows 10
 - AeroFTP Version: v4.0.4

**Additional context**
It looks like my description in https://github.com/axpdev-lab/aeroftp/discussions/270#discussioncomment-17288728 and may relate to it. Something tells me that the issue of the Cryptomater Vault window being dark is likely due to not using it for a while, forgetting about it, and not updating it.
Some update along the way likely caused Cryptomater Vault to stop working. That's my guess.

axpnet:
This is fixed in v4.0.5.
Two separate defects fed into it, both now resolved. The unlock window forced a manual folder re-pick and the masterkey lookup then ran against the wrong directory, which is where the "cannot find masterkey.cryptomator" error came from. v4.0.5 prefills the correct vault root from the file you right-clicked, so there is nothing to re-pick. A second defect (a crypto feature flag) was breaking Cryptomator vault creation and GitHub App auth and is fixed in the same release. Thanks for the precise report and the linked context.

EhudKirsh:
I tested this feature now twice in v4.0.7 and it still doesn't work, but now it fails silently instead of by throwing an error. 
I go to the `NewVault` folder that it creates, and right-click on `vault.cryptomator` and `masterkey.cryptomator` and then click on
'🔒 Open as Cryptomator Vault'. I enter the password, and the failure it shows is that the vault is empty, even though I created it by selecting items, right-clicking on them and then clicking '🔒 Create Cryptomator Vault...'.

axpnet:
Hi Ehud, thank you for the careful retests and for pinning down the "opens empty" symptom. You were right, and we owe you a precise explanation.

[...]

Issue #215, opened 2026-05-17
[BUG v3.8.2] Can't switch protocols when editing Filen profiles


EhudKirsh:
**Describe the bug**
I really like the new 3 radio buttons in the Filen Quick Connect page, that allow manually switching between API, WebDAV and S3. But it seems to glitch out in v3.8.2 though. Please try to edit an existing Filen API profile and click on the WebDAV or S3 options. Now these 3 options disappear for some reason. I'd like to see them all stay to be able to switch back and forth. Now trying to edit an existing Filen S3 profile doesn't let me switch to WebDAV. Clicking on WebDAV does nothing, even if switching to API and then WebDAV. Same thing, trying to edit Filen WebDAV and trying to switch to S3, doesn't let switching to it when clicking on it, even if switching to API and then trying to click on S3.

**Steps To Reproduce**
As I described above, trying to switch existing Filen profiles to other protocols often doesn't work.

**Expected behavior**
I expect to be able to edit an existing Filen profile and switch it from any of the 3 protocols to any other of them.

**Environment:**
 - OS: Windows 10
 - AeroFTP Version: v3.8.2

**Additional context**
- This doesn't seem to be an issue with setting Filen up, starting from any of these 3. It's only with existing Filen profiles.
- I'm pretty sure that this wasn't the case in v3.8.1, so I suggest that after you go through the steps to see the problem for yourself in v3.8.2, that you test this in v3.8.1 and maybe also v3.8.0 and v3.7.9 (I don't remember when this was introduced), because I think that some of them might've worked fine. I guess this is a good way to debug in general, look for an older version when things worked and see what went wrong during some update. It should be easier to debug than a bug that always existed with a feature and there's no version when a feature worked without a bug.

axpnet:
Root-caused. On v3.8.2 the three Filen mode buttons were treated as incompatible protocols while editing a saved profile, so selecting another dropped you out of edit mode and the strip vanished.

v3.8.3 fixes the disappearing: the mode strip now stays visible while you edit a saved profile. Please retest on v3.8.3.

[...]

Issue #128, opened 2026-04-25
[FEATURE] Add a '2FA Code' input field when connecting MEGA


EhudKirsh:
**Is your feature request related to a problem?**
In the 'Discover Services' page there are 3 E2E cloud drives: MEGA, Filen and Internxt. Clicking on Filen and Internxt shows a '2FA Code' input field, but not when clicking on MEGA, despite the fact that MEGA uses 2FA TOTP as well. I think this field might've been forgotten when the MEGA 'Quick Connect' page was built and more focus was likely put on the Native API VS MEGAcmd.

**Describe the solution you'd like**
I don't know if 2FA is needed to connect to MEGAcmd, but if so, a good place to place this '2FA Code' field would be under Password and above CONNECTION BACKEND, since both the Native API and MEGAcmd would rely on this 2FA Code like they both rely on the Email Account and Password.

**Describe alternatives you've considered**
Not sure what alternatives there could be. Adding a '2FA Code' input field when connecting MEGA like Filen and Internxt have is a pretty simple and elegant solution I think, the only one needed.

**Additional context**
N/A

axpnet:
Tracked for one of the next releases. The MEGA backend currently has no 2FA plumbing at all (Filen and Internxt do, MEGA does not), so this is not a UI-only patch: it needs the MEGA login path extended to forward the TOTP code through the SRP exchange. The UI side will mirror the Filen/Internxt block once the backend is ready.

Thanks for spotting it, Leaving this open until it ships.

👋 ciao

EhudKirsh:
More E2E feedback:

I noticed that merely connecting Filen and waiting a few minutes gives this error when trying to see the files and folders stored in Filen (i.e. clicking on the Filen icon in the 'My Servers' page):

    ❌ Connection failed
    Connection failed: Authentication failed: Wrong
    Two Factor Authentication code.

You can see when editing Filen (right-click > '🖊 Edit') that the 2FA Code text field just writes there the 6-digit TOTP from the last connection and nothing else.

There's got to be a way to persist E2E. I suggest you have a look at how rclone.conf handles MEGA's 2FA TOTP. I think the solution is likely there. I looked at my rclone.conf unencrypted and it looks something like this:

```txt
[MEGA]
type = mega
2fa = 123456 (not the exact TOTP, just an example of the 6-digits there)
pass = (a long string)
user = (my email address 📧)
session_id = (a long string)
master_key = (a short string)
```

[...]

PR #875, opened 2026-09-19
fix: S3 bucket glyph, OAuth and 2FA edit Save, Crypt reveal, All-tab tiers, AeroCrypt v4 sizes

axpnet:
Five items from a review round on #347, #369, #215 and #274, each in its own commit.

**S3 draws a bucket** (#347). S3 drew lucide `Database`, a shape S3 has no concept for. The bucket was asked for and first declined to keep the #567 shape; that decision is reversed. lucide ships no bucket (`PaintBucket` is the fill tool), so a `Bucket` glyph is drawn on lucide's grid and stroke and added to the shared connection-method map. The map was half the job: the S3 tab in the provider selector, the My Servers sidebar filter, the session tab, the S3 header in Quick Connect, the Object Storage category and the four S3 presets each drew `Database` on their own. All eight now draw the bucket, and the Azure Blob header, which also borrowed `Database`, draws the Blob container. A test fails if any of those surfaces draws `Database` again; it fails on `main`.

**The OAuth edit Save persists overlay changes; the 2FA Save stays enabled with a code typed** (#369). The OAuth edit Save (#400) only enabled on a name, path or icon change, so changing a Wrappers/Overlays setting left it greyed out and a new sign-in was the only way to persist it. `oauthEditHasChanges` now also compares the overlay the way the save writes it (`overlayEditDiffers`, a pure function with its own tests), and the metadata save, which bypasses `saveToServers`, gets the three overlay gates `saveToServers` already applies (#322 confirm mismatch, #369 invalid Overlays Remote Path, #276 default-salt entropy). On the 2FA pages the edit Save disabled itself while a TOTP code was typed; `saveToServers` never persists a code (#128), so Save now stays enabled and drops the code, and the tooltip still says how to apply it now.

**Show and copy a stored Crypt password and salt** (#215). On a bound overlay the fields are locked and open empty, since the secrets live in the vault, so the eye toggled the mask of an empty field. The eye now reads the stored secret on demand, race-guarded like the other vault reads, and shows it read-only with a copy button. The revealed value lives in its own state, so revealing never re-stores it or counts as an edit, and the copy tooltip never carries the secret.

**A company appears under All with every tier it has in any tab** (#274). With Paid selected, MEGA appeared under the S3 tab but not under All: a tab classifies by the methods in that category, while All used the whole-company tier. All now takes the union. The test is the one suggested on #274, run over the whole catalog: no company may appear in a tab and tier and be missing from All with that tier. On `main` it fails with three rows, not one: MEGA under S3, pCloud Drive and Yandex Disk under WebDAV.

**AeroCrypt sizes map for v4 vaults, and in the CLI and MCP compare** (#347). A vault migrated to keyslots keeps the v3 object codec, but both size callers matched `V3` only or not at all: the transparent overlay reported a v4 vault's ciphertext size as inexact, so a GUI Compare dropped the size check, and `check`, `reconcile` and `check_tree` mapped no AeroCrypt size, so every AeroCrypt file read as differing. `overlay::config_decrypted_size` is now the one place that decides it. The tests use real objects from the codec, and three of them fail with the map narrowed back to `V3`.

## Verification

- `npm run typecheck`: 0 errors. `npm run test:unit`: 107 files, 942 tests, all passing.
- `cargo fmt --check` clean; `cargo test --lib` on the crypt size tests: 16 passing; `cargo clippy --lib --tests -D warnings`: clean after `cargo clean -p aeroftp`. The CLI binary is untouched and left to CI.
- Each new test was seen failing without its fix: the glyph surfaces (one surface reverted), the catalog union (union removed), the v4 size map (map narrowed to `V3`).

## Not verified

- None of the UI changes has been exercised in the running app. The overlay Save, the TOTP Save, the Crypt reveal and the bucket rendering are covered by type checks, unit tests on the pure parts and reading the code, not by a click.
- Legacy AeroCrypt v1/v2 overlays keep the deferred size behaviour, as before.
- The #347 Compare report may also concern an rclone Crypt profile on a recent version; that case is not reproduced here and the reporter is being asked which version and kind it was.


axpnet:
@coderabbitai full review

